PR #4, opened 2024-01-09
removed faulty code on swerve init

Lewis-Seiden:


Lewis-Seiden:
less clean, but the casts didnt work and i forgot to sim them when i wrote them

PR #6, opened 2024-01-13
added robot init unit test

Lewis-Seiden:


src/test/java/frc/InitTests.java
```diff
@@ -0,0 +1,25 @@
...
+
+/** Test robot init. */
+public class InitTests {
+  @BeforeEach
+  private void setup() {
+    HAL.initialize(500, 0);
+  }
+
+  @Test
+  private void initTest() {
```
kevinclark, line 20: Prefer test names that say what is expected. These are never called, so you should be excessively verbose since it's what's going to communicate intent when it fails. F.e, "robotShouldNotCombustWhenFedFrechfries()" would be a totally reasonable unit test method name.

kevinclark: Also note you don't event need test in the name - you've already communicated it's a test with the annotation

kevinclark: All that being said, what's your goal with this method? When does it fail? What does it tell you when it does?

Awesomeplayer165 (commented):
LGTM!

.vscode/settings.json
```diff
@@ -26,5 +26,6 @@
     },
   ],
   "java.test.defaultConfig": "WPIlibUnitTests",
-  "java.compile.nullAnalysis.mode": "automatic"
+  "java.compile.nullAnalysis.mode": "automatic",
+  "wpilib.skipTests": true
```
kevinclark, line 30: Is this set to true on purpose?

kevinclark (commented):
Unclear this is useful yet and expect it's in progress. Some notes though.

src/test/java/frc/InitTests.java
Lewis-Seiden: the goal here is just to make sure `Robot` initializes without throwing an error, mainly as a simple template for unit tests as we (hopefully) add more throughout the season

.vscode/settings.json
Lewis-Seiden: yes, to avoid excessive delays during deploys from running unit tests, and instead my plan was to run these tests in ci

src/test/java/frc/InitTests.java
kevinclark: Then I'd make it explicit with a [assertDoesNotThrow](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)). Communicates your intent and makes clear you didn't forget the assert along the way.

PR #25, opened 2024-01-19
Vision

spellingcat:
yippee!


src/main/java/frc/robot/Robot.java
```diff
@@ -38,6 +41,7 @@ public static enum RobotMode {
   private final SwerveSubsystem swerve =
       new SwerveSubsystem(
           mode == RobotMode.REAL ? new GyroIOPigeon2() : new GyroIO() {},
+          mode == RobotMode.REAL ? new VisionIOReal() : new VisionIOSim(),
```
Lewis-Seiden, line 43: probably should have a way to input multiple visionio instances, maybe make it an array in the constructor. if you do that the modules should also be an array for consistency

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -0,0 +1,25 @@
...
+/** Add your docs here. */
+public interface VisionIO {
+
+  public static class VisionIOInputs {
+    public double timestamp = 0.0;
+    public double latency = 0.0;
+    public List<PhotonTrackedTarget> targets =
+        new ArrayList<>(); // TODO make protobuf work whenever that happens
+    public double numTags = 0;
+    public Pose3d pose = new Pose3d();
```
Lewis-Seiden, line 21: clarify where this is ocming from. if it can be done above the io layer that would be great. if its from the coproc clarify

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.photonvision.targeting.PhotonTrackedTarget;
+
+/** Add your docs here. */
+public interface VisionIO {
+
+  public static class VisionIOInputs {
+    public double timestamp = 0.0;
+    public double latency = 0.0;
```
Lewis-Seiden, line 17: are these redundant?

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -0,0 +1,25 @@
...
+    public List<PhotonTrackedTarget> targets =
+        new ArrayList<>(); // TODO make protobuf work whenever that happens
+    public double numTags = 0;
+    public Pose3d pose = new Pose3d();
+  }
+
+  public default void updateInputs(VisionIOInputs inputs, Pose3d pose) {};
```
Lewis-Seiden: do you always need the pose or just in sim?

src/main/java/frc/robot/subsystems/Vision/VisionIOInputsLogged.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+    table.put("NumTags", numTags);
+    table.put("Pose", pose);
+  }
+
+  @Override
+  public void fromLog(LogTable table) {
+    timestamp = table.get("Timestamp", timestamp);
+    latency = table.get("Latency", latency);
+    for (int i = 0; i < table.get("number of tags", targets.size()); i++) {
```
Lewis-Seiden, line 28: is this the right label?

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -0,0 +1,25 @@
...
+
+/** Add your docs here. */
+public interface VisionIO {
+
+  public static class VisionIOInputs {
+    public double timestamp = 0.0;
+    public double latency = 0.0;
+    public List<PhotonTrackedTarget> targets =
+        new ArrayList<>(); // TODO make protobuf work whenever that happens
+    public double numTags = 0;
```
Lewis-Seiden, line 20: does this need to be in inputs or just in the table

src/main/java/frc/robot/subsystems/Vision/VisionIOInputsLogged.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+    numTags = table.get("NumTags", numTags);
+    pose = table.get("Pose", pose);
+  }
+
+  public VisionIOInputsLogged clone() {
+    VisionIOInputsLogged copy = new VisionIOInputsLogged();
+    copy.timestamp = this.timestamp;
+    copy.latency = this.latency;
+    copy.targets = this.targets;
```
Lewis-Seiden, line 39: make sure to copy targets if you want to deep copy

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
```diff
@@ -0,0 +1,23 @@
...
+
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Pose3d;
+import org.photonvision.PhotonCamera;
+
+/** Add your docs here. */
+public class VisionIOReal implements VisionIO {
+  // constants
+  public static final String CAMERA_NAME = "opi-camera";
```
Lewis-Seiden, line 13: this shouldnt be hardcoded, pass it through the constructor

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
```diff
@@ -0,0 +1,23 @@
...
+  public static PhotonCamera camera = new PhotonCamera(CAMERA_NAME);
+
+  @Override
+  public void updateInputs(VisionIOInputs inputs, Pose3d pose) {
+    var result = camera.getLatestResult();
```
Lewis-Seiden, line 18: make sure to update all of inputs fields

src/main/java/frc/robot/subsystems/Vision/VisionIOSim.java
```diff
@@ -0,0 +1,38 @@
...
+import edu.wpi.first.apriltag.AprilTagFieldLayout.OriginPosition;
+import edu.wpi.first.apriltag.AprilTagFields;
+import edu.wpi.first.math.geometry.Pose3d;
+import org.photonvision.PhotonCamera;
+import org.photonvision.simulation.VisionSystemSim;
+
+/** Add your docs here. */
+public class VisionIOSim implements VisionIO {
+  public static final String SIM_VISION_SYSTEM_NAME = "sim vision";
+  public static final String SIM_CAMERA_NAME = "sim camera";
```
Lewis-Seiden, line 16: make sure to have diff cameras have diff names

src/main/java/frc/robot/subsystems/Vision/VisionIOSim.java
```diff
@@ -0,0 +1,38 @@
...
+      sim.addAprilTags(field);
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  @Override
+  public void updateInputs(VisionIOInputs inputs, Pose3d pose) {
+    var result = simCamera.getLatestResult();
+    sim.update(pose);
```
Lewis-Seiden, line 33: make sure to set the cameras pose right

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -68,8 +81,18 @@ public class SwerveSubsystem extends SubsystemBase {
   private Pose2d pose = new Pose2d();
   private Rotation2d lastGyroRotation = new Rotation2d();
 
-  public SwerveSubsystem(GyroIO gyroIO, ModuleIO... moduleIOs) {
+  private final VisionIO visionIO;
+  // private final VisionIOInputsAutoLogged visionInputs = new VisionIOInputsAutoLogged();
+  private final VisionIOInputsLogged visionInputs = new VisionIOInputsLogged();
```
Lewis-Seiden, line 85: make sure to support multiple visionios and remove the commented out code

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -178,7 +206,83 @@ public void periodic() {
       }
       // Apply the twist (change since last sample) to the current pose
       pose = pose.exp(twist);
+      estimator.update(lastGyroRotation, modulePositions); // FIXME sort of sketch
```
Lewis-Seiden, line 208: also check 6328s pose estimator, it might be cleaner than this

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -178,7 +206,83 @@ public void periodic() {
...
     }
+    Logger.recordOutput("Swerve Pose", pose);
+
+    visionIO.updateInputs(visionInputs, new Pose3d(pose));
```
Lewis-Seiden, line 212: update inputs at the top of periodic

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -178,7 +206,83 @@ public void periodic() {
...
+              .estimatedPose;
+      Logger.recordOutput("Vision Pose", visionPose);
+      estimator.addVisionMeasurement(
+          visionPose.toPose2d(),
+          visionInputs.timestamp,
+          VisionHelper.findVisionMeasurements(estPose));
+    } catch (NoSuchElementException e) {
+    }
+
+    // From 6995 - untested
```
Lewis-Seiden, line 241: remove commented out code

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -178,7 +206,83 @@ public void periodic() {
...
+    //         }
+    //         estimator.addVisionMeasurement(
+    //                 measurement.estimation().estimatedPose.toPose2d(),
+    //                 measurement.estimation().timestampSeconds,
+    //                 measurement.confidence());
+    //         Logger.recordOutput("Odo + Vision Pose", estimator.getEstimatedPosition());
+    //       }
+    //     } catch (NoSuchElementException e) {
+    //     }
+    Logger.recordOutput(
```
Lewis-Seiden, line 280: this can be removed bc getPose exists

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+import org.photonvision.PhotonPoseEstimator.PoseStrategy;
+import org.photonvision.estimation.TargetModel;
+import org.photonvision.estimation.VisionEstimation;
+import org.photonvision.targeting.PhotonPipelineResult;
+import org.photonvision.targeting.PhotonTrackedTarget;
+import org.photonvision.targeting.TargetCorner;
+
+public class VisionHelper {
+  // constants
+  public static final Matrix<N3, N3> CAMERA_MATRIX_OPT =
```
Lewis-Seiden, line 37: make sure to have different matrices for each camera

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+  public static final Matrix<N5, N1> DIST_COEFFS_OPT =
+      MatBuilder.fill(
+          Nat.N5(),
+          Nat.N1(),
+          0.05452153950284706,
+          -0.04331612051891956,
+          0.00176988756858703,
+          -0.004530368741385627,
+          -0.040501622476628085); // TODO tune maybe
+  public static final Transform3d CAMERA_TO_ROBOT = new Transform3d(); // TODO find
```
Lewis-Seiden, line 59: make sure to have different transforms for each camera

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+    double poseCacheTimestampSeconds = -1;
+
+    // Time in the past -- give up, since the following if expects times > 0
+    if (result.getTimestampSeconds() < 0) {
+      return Optional.empty();
+    }
+
+    // If the pose cache timestamp was set, and the result is from the same timestamp, return an
+    // empty result
+    // if (poseCacheTimestampSeconds > 0
```
Lewis-Seiden, line 191: deal w commented out code

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+      MatBuilder.fill(
+          Nat.N5(),
+          Nat.N1(),
+          0.05452153950284706,
+          -0.04331612051891956,
+          0.00176988756858703,
+          -0.004530368741385627,
+          -0.040501622476628085); // TODO tune maybe
+  public static final Transform3d CAMERA_TO_ROBOT = new Transform3d(); // TODO find
+  public static final List<TagCountDeviation> TAG_COUNT_DEVIATION_PARAMS =
```
Lewis-Seiden, line 60: could you add a doc comment to this

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+    logTransform3d(target.getBestCameraToTarget(), table, name);
+    logTransform3d(target.getAlternateCameraToTarget(), table, "alt " + name);
+    logCorners(target, table, name);
+
+    table.put("yaw " + name, target.getYaw());
+    table.put("pitch " + name, target.getPitch());
+    table.put("area " + name, target.getArea());
+    table.put("skew " + name, target.getSkew());
+    table.put("fiducial id " + name, target.getFiducialId());
+    table.put("pose ambiguity " + name, target.getPoseAmbiguity());
```
Lewis-Seiden, line 88: probably should have consistent capitalization schemes with the other logged fields but tbh doesnt matter

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+        table.get("skew " + name, -1),
+        (int) (table.get("fiducial id " + name, -1)),
+        pose,
+        altPose,
+        table.get("pose ambiguity " + name, -1),
+        minAreaRectCorners,
+        detectedCorners));
+  }
+  /**
+   * Poll data from the configured cameras and update the estimated position of the robot. Returns
```
Lewis-Seiden, line 170: update this comment

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+      default:
+        DriverStation.reportError(
+            "[PhotonPoseEstimator] Unknown Position Estimation Strategy!", false);
+        return Optional.empty();
+    }
+
+    return estimatedPose;
+  }
+
+  private static Optional<EstimatedRobotPose> multiTagPNPStrategy(
```
Lewis-Seiden, line 222: clarify that this is pnp on rio, maybe make sure it uses pnp on coproc when available

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+              result.targets,
+              fieldTags,
+              TargetModel.kAprilTag36h11);
+      var best =
+          new Pose3d()
+              .plus(pnpResults.best) // field-to-camera
+              .plus(CAMERA_TO_ROBOT); // field-to-robot
+      // var alt = new Pose3d()
+      // .plus(pnpResults.alt) // field-to-camera
+      // .plus(robotToCamera.inverse()); // field-to-robot
```
Lewis-Seiden, line 282: clean up dead code

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+    double avgDistance = sumDistance / estimation.targetsUsed.size();
+
+    var deviation =
+        TAG_COUNT_DEVIATION_PARAMS
+            .get(
+                MathUtil.clamp(
+                    estimation.targetsUsed.size() - 1, 0, TAG_COUNT_DEVIATION_PARAMS.size() - 1))
+            .computeDeviation(avgDistance);
+
+    // System.out.println(
```
Lewis-Seiden, line 392: clean up dead code

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,423 @@
...
+    //         "with %d tags at smallest distance %f and pose ambiguity factor %f, confidence
+    // multiplier %f",
+    //         estimation.targetsUsed.size(),
+    //         smallestDistance,
+    //         poseAmbiguityFactor,
+    //         confidenceMultiplier));
+
+    return deviation;
+  }
+  // Reject unreasonable vision poses
```
Lewis-Seiden, line 403: this should be a doc comment

Lewis-Seiden (approved):

src/main/java/frc/robot/Robot.java
spellingcat: can i do the thing with the ellipses in the constructor

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: wdym where is it coming from, it should all be running on the coproc

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: I don't think so, timestamp is like when in the match it happened and latency should just be latency

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: i think just sim, but i still need to figure that out

src/main/java/frc/robot/subsystems/Vision/VisionIOInputsLogged.java
spellingcat: no it's not

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: should be in inputs i think?? if i'm wrong correct me

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
spellingcat: which constructor?

src/main/java/frc/robot/subsystems/Vision/VisionIOSim.java
spellingcat: will do once i get sim figured out

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: by "this" do you mean line 281

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
spellingcat: tbqh this is from 5026 and i am not entirely sure how it works

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
Lewis-Seiden: ye just put that in the variable name

Lewis-Seiden: and maybe also that its from multi tag pnp

src/main/java/frc/robot/subsystems/Vision/Vision.java
```diff
@@ -0,0 +1,55 @@
...
+          923.5403619629557,
+          0.0,
+          644.4965658066068,
+          0.0,
+          925.8136962361125,
+          402.6412935350414,
+          0.0,
+          0.0,
+          1.0); // TODO tune maybe
+  public final Matrix<N5, N1> DIST_COEFFS_OPT =
```
Lewis-Seiden, line 33: these should be part of VisionConstants bc they are per lens

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -15,12 +15,11 @@ public interface VisionIO {
   public static class VisionIOInputs {
     public double timestamp = 0.0;
     public double latency = 0.0;
-    public List<PhotonTrackedTarget> targets =
+    public List<PhotonTrackedTarget> coprocPNPTargets = // TODO is this a bad name
```
Lewis-Seiden, line 18: i mean this isnt coproc pnp, this is just targets

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
```diff
@@ -15,12 +15,11 @@ public interface VisionIO {
...
         new ArrayList<>(); // TODO make protobuf work whenever that happens
-    public double numTags = 0;
+    public double numTags = 0; // TODO why isn't this just targets.size()?
     public Pose3d pose = new Pose3d();
```
Lewis-Seiden, line 21: the pose is the coproc pnp

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
```diff
@@ -5,19 +5,38 @@
...
+
+  /*** Transform3d from the center of the robot to the camera mount position (ie,
+   *     robot ➔ camera) in the <a href=
+   *     "https://docs.wpilib.org/en/stable/docs/software/advanced-controls/geometry/coordinate-systems.html#robot-coordinate-system">Robot
+   *     Coordinate System</a>.
+   ***/
+  public Transform3d robotToCamera;
+
+  public VisionIOReal(VisionConstants constants) {
+    CAMERA_NAME = constants.cameraName();
```
Lewis-Seiden, line 26: not a constant bc you set it here

src/main/java/frc/robot/subsystems/Vision/VisionIOSim.java
```diff
@@ -4,41 +4,46 @@
...
-
 /** Add your docs here. */
 public class VisionIOSim implements VisionIO {
-  public static final String SIM_VISION_SYSTEM_NAME = "sim vision";
-  public static final String SIM_CAMERA_NAME = "sim camera";
-  VisionSystemSim sim = new VisionSystemSim(SIM_VISION_SYSTEM_NAME);
-  PhotonCamera camera = new PhotonCamera(SIM_CAMERA_NAME); //TODO ???? what how why
-  SimCameraProperties cameraProp = new SimCameraProperties(); //TODO find
+  private final String SIM_VISION_SYSTEM_NAME;
+  private final String SIM_CAMERA_NAME;
```
Lewis-Seiden, line 19: not a constant

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -80,23 +84,30 @@ public class SwerveSubsystem extends SubsystemBase {
...
+  private final Vision[] cameras;
   private AprilTagFieldLayout fieldTags;
   public SwerveDrivePoseEstimator estimator;
   Vector<N3> odoStdDevs = VecBuilder.fill(0.3, 0.3, 0.01);
   Vector<N3> visStdDevs = VecBuilder.fill(1.3, 1.3, 3.3);
   private SwerveModulePosition[] modulePositions = new SwerveModulePosition[4];
-
-  public SwerveSubsystem(GyroIO gyroIO, VisionIO visionIO, ModuleIO... moduleIOs) {
+  public static final Transform3d leftCamToRobot = new Transform3d(); // TODO find
+  public static final Transform3d rightCamToRobot = new Transform3d(); // TODO find
```
Lewis-Seiden, line 94: these transforms shouldnt need to be in their own variables and can just be defined in the constant

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -80,23 +84,30 @@ public class SwerveSubsystem extends SubsystemBase {
...
+  public static final VisionConstants leftCam = new VisionConstants("Left Camera", leftCamToRobot);
+  public static final VisionConstants rightCam =
+      new VisionConstants("Right Camera", rightCamToRobot);
+  public static final VisionConstants[] cameraConstants = new VisionConstants[] {leftCam, rightCam};
```
Lewis-Seiden, line 98: dont need to store this i think, just have it in a method like the modules

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -80,23 +84,30 @@ public class SwerveSubsystem extends SubsystemBase {
...
+
+  public SwerveSubsystem(GyroIO gyroIO, VisionIO[] visionIOs, ModuleIO... moduleIOs) {
```
Lewis-Seiden, line 100: change moduleios to jsut be an array instead of using `...` for consistency

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -147,8 +158,32 @@ public static ModuleIO[] createSimModules() {
       new ModuleIOSim("BackRight")
     };
   }
+  /**
+   * Constructs an array of vision IOs corresponding to the real robot.
+   *
+   * @return The array of vision IOs.
+   */
+  public static VisionIO[] createCameras() {
```
Lewis-Seiden, line 166: name `createRealCameras` to distinguish from sim

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -147,8 +158,32 @@ public static ModuleIO[] createSimModules() {
...
+  public static VisionIO[] createSimCameras() {
+    return new VisionIO[] {
+      new VisionIOSim("LeftCam", "LeftSystem"), // TODO
+      new VisionIOSim("RightCam", "RightSystem")
+    };
+  }
 
   public void periodic() {
+    for (var camera : cameras) {
+      camera.io.updateInputs(camera.inputs, new Pose3d(pose)); // TODO so sketch
```
Lewis-Seiden, line 184: not sketch i think

src/main/java/frc/robot/subsystems/Vision/VisionIOSim.java
spellingcat: sim.addCamera does that

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: ?? so what does [this](https://github.com/HighlanderRobotics/Crescendo/pull/25#discussion_r1458111913) mean then

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: oh mb

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
```diff
@@ -23,20 +29,20 @@ public class VisionIOReal implements VisionIO {
...
-  public void updateInputs(VisionIOInputs inputs, Pose3d pose) { // TODO
+  public void updateInputs(VisionIOInputs inputs, Pose3d pose) {
     var result = camera.getLatestResult();
     inputs.timestamp = result.getTimestampSeconds();
     inputs.latency = result.getLatencyMillis();
-    inputs.coprocPNPTargets = result.targets;
+    inputs.targets = result.targets;
     inputs.numTags = result.targets.size();
-    inputs.pose = pose;
+    inputs.coprocPNPPose = pose;
```
Lewis-Seiden, line 46: dont just pass through the input pose here

src/main/java/frc/robot/subsystems/swerve/ModuleIO.java
```diff
@@ -32,6 +32,9 @@ public static class ModuleIOInputs {
 
     public double[] odometryDrivePositionsMeters = new double[] {};
     public Rotation2d[] odometryTurnPositions = new Rotation2d[] {};
+    public double[] odometryTimestamps = new double[] {};
+    public double[] odometryDrivePositionsRad =
+        new double[] {}; // TODO is this diff from odometryDrivePositionsMeters
```
Lewis-Seiden, line 37: i dont think so, pretty sure i just changed the units in the swerve cleanup pr

src/main/java/frc/robot/subsystems/swerve/ModuleIOTalonFX.java
```diff
@@ -198,12 +202,27 @@ public void updateInputs(ModuleIOInputs inputs) {
...
+        timestampQueue.stream().mapToDouble((Double value) -> value).toArray();
+    inputs.odometryDrivePositionsRad =
+        drivePositionQueue.stream()
+            .mapToDouble(
+                (Double value) -> Units.rotationsToRadians(value) / Module.DRIVE_GEAR_RATIO) // TODO
+            .toArray();
     inputs.odometryTurnPositions =
         turnPositionQueue.stream()
-            .map(Rotation2d::fromRotations) // should be after offset + gear ratio
+            .map((Double value) -> Rotation2d.fromRotations(value / Module.TURN_GEAR_RATIO))
```
Lewis-Seiden, line 223: all of our readings from the modules should be in meters since the motors are configured w a mechanism ratio thingy

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -82,22 +87,81 @@ public class SwerveSubsystem extends SubsystemBase {
...
+          -0.04331612051891956,
+          0.00176988756858703,
+          -0.004530368741385627,
+          -0.040501622476628085); // TODO find!!
+  public static final VisionConstants leftCam =
+      new VisionConstants(
+          "Left Camera",
+          new Transform3d(),
+          "Left Camera Sim System",
+          LEFT_CAMERA_MATRIX_OPT,
```
Lewis-Seiden, line 154: probably should remove the OPT if that stands for optional

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -173,16 +237,13 @@ public static VisionIO[] createCameras() {
...
-    };
+    return new VisionIO[] {new VisionIOSim(leftCam), new VisionIOSim(rightCam)};
   }
 
   public void periodic() {
     for (var camera : cameras) {
-      camera.io.updateInputs(camera.inputs, new Pose3d(pose)); // TODO so sketch
-      Logger.processInputs("Vision", camera.inputs); // TODO so sketch
+      camera.updateInputs(new Pose3d(pose));
+      Logger.processInputs("Vision", camera.inputs);
```
Lewis-Seiden, line 246: probably need each camera to have a different name for the log, could just use the camera name

Lewis-Seiden: or follow modules example and handle the Logger call in Vision.updateInputs

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -208,67 +269,59 @@ public void periodic() {
...
-            Arrays.stream(modules)
-                .map((m) -> m.getPositionDeltas().length)
-                .min(Integer::compare)
-                .get());
-    for (int deltaIndex = 0; deltaIndex < deltaCount; deltaIndex++) {
+    double[] sampleTimestamps =
+        modules[0].getOdometryTimestamps(); // All signals are sampled together
+    int sampleCount = sampleTimestamps.length;
+
+    // int deltaCount =
```
Lewis-Seiden, line 276: remove commented out stuff

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -208,67 +269,59 @@ public void periodic() {
...
-        // with the change in angle since the last sample.
-        Rotation2d gyroRotation = gyroInputs.odometryYawPositions[deltaIndex];
-        twist = new Twist2d(twist.dx, twist.dy, gyroRotation.minus(lastGyroRotation).getRadians());
-        lastGyroRotation = gyroRotation;
+        rawGyroRotation = gyroInputs.odometryYawPositions[deltaIndex];
+      } else {
+        // Use the angle delta from the kinematics and module deltas
+        Twist2d twist = kinematics.toTwist2d(wheelDeltas);
+        rawGyroRotation = rawGyroRotation.plus(new Rotation2d(twist.dtheta));
+        pose = pose.exp(twist);
```
Lewis-Seiden, line 303: shouldnt this be called no matter if the gyro is connected or not?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -208,67 +269,59 @@ public void periodic() {
...
-                    result,
-                    camera.CAMERA_MATRIX_OPT,
-                    camera.DIST_COEFFS_OPT,
-                    PoseStrategy.MULTI_TAG_PNP_ON_COPROCESSOR,
-                    fieldTags,
-                    camera.constants.robotToCamera())
-                .get()
-                .estimatedPose;
+        var visionPose = estPose.estimatedPose;
         Logger.recordOutput("Vision Pose", visionPose);
```
Lewis-Seiden, line 325: make sure this handles having multiple vision estimates in one loop

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -428,7 +481,7 @@ public Rotation2d getRotation() {
     return pose.getRotation();
   }
 
-  /** Resets the current odometry pose. */
+  /** Sets the current odometry pose. */
   public void setPose(Pose2d pose) {
```
Lewis-Seiden, line 485: does this reset the pose estimator?

Lewis-Seiden (changes requested):

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
```diff
@@ -0,0 +1,428 @@
...
+        break;
+      case MULTI_TAG_PNP_ON_COPROCESSOR:
+        estimatedPose =
+            multiTagOnCoprocStrategy(
+                cameraResult,
+                Optional.of(cameraMatrix),
+                Optional.of(distCoeffs),
+                fieldTags,
+                robotToCamera,
+                strat); // TODO make multiple cameras/coprocs actually work
```
Lewis-Seiden, line 203: make sure to pass this a non pnp strat

Lewis-Seiden (changes requested):

src/main/java/frc/robot/Robot.java
Lewis-Seiden: youve prob figured it out by now but you can only do that for the last item in the constructor

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
Lewis-Seiden: if its just sim maybe make the sim io constructor take in a `Supplier<Pose3d>` which it calls instead of having `updateInputs` always take it

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
Lewis-Seiden: idt it needs to but tbh its fine

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
Lewis-Seiden: yes since the swerve pose is alr logged w `@AutoLogOutput`

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
Lewis-Seiden: might be worth dming isaac

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
Lewis-Seiden: also make sure that the actual fields in the record are formatted properly

src/main/java/frc/robot/subsystems/Vision/VisionIO.java
spellingcat: but then how do i get it to update every cycle? (sorry if this is really obvious and i'm just missing something)

src/main/java/frc/robot/subsystems/Vision/VisionHelper.java
spellingcat: will do at some point

src/main/java/frc/robot/subsystems/Vision/Vision.java
```diff
@@ -15,32 +13,14 @@
 
 /** Add your docs here. */
 public class Vision {
-  public record VisionConstants(String cameraName, Transform3d robotToCamera) {}
+  public record VisionConstants(
+      String cameraName,
+      Transform3d robotToCamera,
+      String simVisionSystemName,
```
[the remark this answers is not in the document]
spellingcat: if you mean simVisionSystemName it can probably be the same but i would rather it be clear that the sim system isnt the camera

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
spellingcat: wait wdym

Lewis-Seiden: `pose` is the input pose, or the previously estimated pose. you want coprocPNPPose to be the, well coprocPNPPose

src/main/java/frc/robot/subsystems/Vision/Vision.java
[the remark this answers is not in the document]
Lewis-Seiden: maybe name it `cameraNameSim`

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: i don't think so because it seems like a replacement for the gyro

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: it does not, should it?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -305,71 +304,72 @@ public void periodic() {
...
+                      camera.constants.robotToCamera())
+                  .get();
+          var visionPose = estPose.estimatedPose;
+          visionPoses.add(visionPose);
+          Logger.recordOutput(
+              "Vision/Vision Pose From " + camera.constants.cameraName(), visionPose);
+          estimator.addVisionMeasurement(
+              visionPose.toPose2d(),
+              camera.inputs.timestamp,
+              VisionHelper.findVisionMeasurements(estPose));
```
Lewis-Seiden, line 331: this could be named better, should include StdDeviations or something along those lines

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -305,71 +304,72 @@ public void periodic() {
...
     }
-    Logger.recordOutput("Swerve Pose", pose);
 
-    for (var camera : cameras) {
-      PhotonPipelineResult result =
-          new PhotonPipelineResult(camera.inputs.latency, camera.inputs.targets);
-      result.setTimestampSeconds(camera.inputs.timestamp);
-      try {
-        var estPose =
+    if (Robot.isReal()) {
```
Lewis-Seiden, line 308: is there a way to contain the sim logic to VisionIOSim?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
Lewis-Seiden: yes

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -484,6 +488,10 @@ public Rotation2d getRotation() {
   /** Sets the current odometry pose. */
   public void setPose(Pose2d pose) {
     this.pose = pose;
+    estimator.resetPosition(
+        pose.getRotation(),
+        lastModulePositions,
+        pose); // this might be bad if lastModulePositions hasnt been set yet
```
Lewis-Seiden, line 494: could include an if or a try catch to check for that

Lewis-Seiden (changes requested):

src/main/java/frc/robot/subsystems/Vision/VisionIOReal.java
```diff
@@ -32,17 +31,16 @@ public VisionIOReal(VisionConstants constants) {
...
 
   @Override
-  public void updateInputs(VisionIOInputs inputs, Pose3d pose) {
+  public void updateInputs(VisionIOInputs inputs) {
     var result = camera.getLatestResult();
     inputs.timestamp = result.getTimestampSeconds();
     inputs.latency = result.getLatencyMillis();
     inputs.targets = result.targets;
     inputs.numTags = result.targets.size();
-    inputs.coprocPNPPose = pose;
```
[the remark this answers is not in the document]
spellingcat: if it's real it doesn't really need the pose right? since that'll be calculated later?

src/main/java/frc/robot/subsystems/swerve/ModuleIOTalonFX.java
```diff
@@ -202,25 +202,13 @@ public void updateInputs(ModuleIOInputs inputs) {
...
-
     inputs.odometryTimestamps =
         timestampQueue.stream().mapToDouble((Double value) -> value).toArray();
-    inputs.odometryDrivePositionsRad =
-        drivePositionQueue.stream()
-            .mapToDouble(
-                (Double value) -> Units.rotationsToRadians(value) / Module.DRIVE_GEAR_RATIO) // TODO
-            .toArray();
+    inputs.odometryDrivePositionsMeters =
+        drivePositionQueue.stream().mapToDouble(Units::rotationsToRadians).toArray();
```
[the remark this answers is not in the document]
spellingcat: i mean this is what it was before

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -86,7 +90,7 @@ public class SwerveSubsystem extends SubsystemBase {
   private final Module[] modules; // FL, FR, BL, BR
 
   private SwerveDriveKinematics kinematics = new SwerveDriveKinematics(getModuleTranslations());
-  private Pose2d pose = new Pose2d();
+  private static Pose2d pose = new Pose2d();
```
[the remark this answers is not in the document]
spellingcat: then how would i make the visionIOSim constructors work?

src/main/java/frc/robot/subsystems/swerve/ModuleIOTalonFX.java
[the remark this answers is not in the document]
Lewis-Seiden: yeah what im saying is it might have been wrong before

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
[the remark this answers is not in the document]
Lewis-Seiden: why do they need it to be static?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: i just changed it but i'm not sure if it makes any sense

src/main/java/frc/robot/subsystems/vision/VisionIO.java
```diff
@@ -22,4 +24,7 @@ public static class VisionIOInputs {
   }
 
   public default void updateInputs(VisionIOInputs inputs) {}
+
+  public default void setSimPose(
```
Lewis-Seiden, line 28: is there a way not to have sim specific functionality in the io interface? if not isok but avoiding it is good 👍

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -305,69 +307,32 @@ public void periodic() {
...
-        result.setTimestampSeconds(camera.inputs.timestamp);
-        double latestTimestamp = result.getTimestampSeconds();
-        boolean newResult = Math.abs(latestTimestamp - lastEstTimestamp) > 1e-5;
-        var simEst =
+    List<Pose3d> visionPoses = new ArrayList<>();
+    for (var camera : cameras) {
+      PhotonPipelineResult result =
+          new PhotonPipelineResult(camera.inputs.latency, camera.inputs.targets);
+      result.setTimestampSeconds(camera.inputs.timestamp);
+      boolean newResult = Math.abs(camera.inputs.timestamp - lastEstTimestamp) > 1e-5;
```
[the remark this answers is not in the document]
spellingcat: yeah this happens for each camera

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -482,11 +470,17 @@ public Rotation2d getRotation() {
 
   /** Sets the current odometry pose. */
   public void setPose(Pose2d pose) {
-    SwerveSubsystem.pose = pose;
+    this.pose = pose;
     try {
       estimator.resetPosition(pose.getRotation(), lastModulePositions, pose);
     } catch (Exception e) {
     }
```
Lewis-Seiden, line 477: why does this need a try catch?

spellingcat: bc of this https://github.com/HighlanderRobotics/Crescendo/pull/25#discussion_r1467247571

Lewis-Seiden: me when i have the memory of a goldfish

src/main/java/frc/robot/subsystems/vision/VisionIO.java
```diff
@@ -0,0 +1,37 @@
...
+    public double latency = 0.0;
+    public List<PhotonTrackedTarget> targets =
+        new ArrayList<>(); // TODO make protobuf work whenever that happens
+    public double numTags = 0; // TODO why isn't this just targets.size()?
+    public Pose3d coprocPNPPose = new Pose3d();
+    public Pose3d[] targetPose3ds;
+    public VisionConstants constants;
+  }
+
+  public default void updateInputs(VisionIOInputs inputs) {}
```
kevinclark, line 29: Unless we've got a good reason to do it we should avoid default implementations doing nothing. It makes it easy to forget to override and then you've got a method doing nothing when you didn't intend that.

The `getName` version is pretty innocuous, but I'd cut it from all the other instances.

src/main/java/frc/robot/subsystems/swerve/ModuleIOTalonFX.java
```diff
@@ -197,12 +201,15 @@ public void updateInputs(ModuleIOInputs inputs) {
     inputs.turnAppliedVolts = turnAppliedVolts.getValueAsDouble();
     inputs.turnCurrentAmps = new double[] {turnCurrent.getValueAsDouble()};
 
+    inputs.odometryTimestamps =
+        timestampQueue.stream().mapToDouble((Double value) -> value).toArray();
     inputs.odometryDrivePositionsMeters =
-        drivePositionQueue.stream().mapToDouble(Units::rotationsToRadians).toArray();
+        drivePositionQueue.stream().mapToDouble((Double value) -> value).toArray();
```
kevinclark, line 207: Had a little more time to look at this - the issue is that `odometryDrivePositionMeters` is a `double[]` vs the `Double[]` you'd get out of just calling `stream().toArray()`. What you've got is *ok* and works because `Double` is implicitly unboxed to `double`. But it looks a little funny because you're passing an identity function to map. You might prefer:

```java
inputs.odometryDrivePositionsMeters = drivePositionQueue.stream().mapToDouble(Double::doubleValue).toArray();
```

It's a little more explicit (converting Double -> double).

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,22 +53,40 @@ public static PhoenixOdometryThread getInstance() {
...
+  public void start() {
+    if (timestampQueues.size() > 0) {
+      super.start();
+    }
   }
 
   public Queue<Double> registerSignal(ParentDevice device, StatusSignal<Double> signal) {
-    isCANFD = CANBus.isNetworkFD(device.getNetwork());
-    Queue<Double> queue = new ArrayBlockingQueue<>(100);
+    Queue<Double> queue = new ArrayDeque<>(100);
```
kevinclark, line 66: This seems like a potentially dangerous change in behavior. The `ArrayBlockingQueue` is size limited and blocks the thread when a 101st item is added. This will never block and is unbounded. I assume this got swapped out because the thread was blocking when we were pulling items from the queue too slowly? That speaks to a problem that should probably be solved (does the max queue size need to be bigger? Do we need to drop samples? Do we need to process faster?) vs being worked around.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,22 +53,40 @@ public static PhoenixOdometryThread getInstance() {
...
     signalsLock.lock();
+    SwerveSubsystem.odometryLock.lock();
```
kevinclark, line 68: Activating a lock in an entirely different class is a violation of separation of concerns and should be avoided. It also produces potential for deadlock. Why does this need to happen?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,22 +53,40 @@ public static PhoenixOdometryThread getInstance() {
...
     } finally {
       signalsLock.unlock();
+      SwerveSubsystem.odometryLock.unlock();
+    }
+    return queue;
+  }
+
+  public Queue<Double> makeTimestampQueue() {
+    Queue<Double> queue = new ArrayDeque<>(100);
+    SwerveSubsystem.odometryLock.lock();
```
kevinclark, line 85: Same note as above.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,22 +53,40 @@ public static PhoenixOdometryThread getInstance() {
...
   }
 
   public Queue<Double> registerSignal(ParentDevice device, StatusSignal<Double> signal) {
-    isCANFD = CANBus.isNetworkFD(device.getNetwork());
-    Queue<Double> queue = new ArrayBlockingQueue<>(100);
+    Queue<Double> queue = new ArrayDeque<>(100);
     signalsLock.lock();
+    SwerveSubsystem.odometryLock.lock();
     try {
+      isCANFD = CANBus.isNetworkFD(device.getNetwork());
```
kevinclark, line 70: This happens once for each signal registered but is a global value, so it's going to overwrite previous values. Is that intended? Does this need to be tracked per signal registered?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -96,9 +116,20 @@ public void run() {
       // Save new data to queues
       SwerveSubsystem.odometryLock.lock();
       try {
+        double timestamp = Logger.getRealTimestamp() / 1e6;
+        double totalLatency = 0.0;
```
kevinclark, line 120: You might prefer:

```java
double totalLatency = Arrays.stream(signals).mapToDouble(s -> s.getTimestamp().getLatency()).sum();
```

src/main/java/frc/robot/subsystems/vision/Vision.java
```diff
@@ -0,0 +1,73 @@
...
+import edu.wpi.first.math.numbers.N3;
+import edu.wpi.first.math.numbers.N5;
+import java.util.Optional;
+import org.littletonrobotics.junction.Logger;
+import org.photonvision.EstimatedRobotPose;
+import org.photonvision.PhotonPoseEstimator.PoseStrategy;
+import org.photonvision.targeting.PhotonPipelineResult;
+
+/** Add your docs here. */
+public class Vision {
```
kevinclark, line 20: Should this `implements VisionIO`? And should the methods be `@Override`?

kevinclark (changes requested):
I've got more of this to read, but there's already actionable stuff and I don't want to keep you waiting. Also, handle your upstream merge.

Will look further later today.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -73,17 +98,108 @@ public class SwerveSubsystem extends SubsystemBase {
...
+          Nat.N3(),
+          923.5403619629557,
+          0.0,
+          644.4965658066068,
+          0.0,
+          925.8136962361125,
+          402.6412935350414,
+          0.0,
+          0.0,
+          1.0); // TODO find!!
```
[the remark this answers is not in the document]
spellingcat: I believe it came from a camera config file somewhere but I will need to double check (hence the todo)

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -174,34 +312,70 @@ public void periodic() {
...
+                getModulePositions()[moduleIndex]
+                        .distanceMeters // smth abt odo positions vs module positions
+                    - lastModulePositions[moduleIndex].distanceMeters,
+                getModulePositions()[moduleIndex].angle);
+        lastModulePositions[moduleIndex] = modulePositions[moduleIndex];
       }
 
-      // The twist represents the motion of the robot since the last
-      // sample in x, y, and theta based only on the modules, without
-      // the gyro. The gyro is always disconnected in simulation.
```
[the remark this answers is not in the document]
spellingcat: yeah not sure why that got deleted

src/main/java/frc/robot/subsystems/vision/VisionIO.java
spellingcat: what should i do instead? i'm not sure how else to call it from the io in Vision.java

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: so this is from 6328's template code, which is what we've been using for the multithread stuff. it seems like they addressed that today [here](https://github.com/Mechanical-Advantage/AdvantageKit/issues/73) and [here.](https://github.com/Mechanical-Advantage/AdvantageKit/commit/a6fc30c01ee1ab523ca06da96308a34d2b3bc5a2) I can use that if you think it looks alright, but I'm not super clear on how/why this works so would you be available to have a more detailed discussion on this sometime?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: I'm not sure, where would I move it to in SwerveSubsystem?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: also not sure but it probably shouldn't be overwritten?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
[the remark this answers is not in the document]
kevinclark: Hokay. Wrap up the TODO before merging.

src/main/java/frc/robot/subsystems/vision/Vision.java
spellingcat: it does not

src/main/java/frc/robot/subsystems/vision/VisionIO.java
kevinclark: I'm not sure what you're asking. `VisionIO` is an interface. If you reference it somewhere, you're saying that the object in question implements that interface and you can call that set of methods on it, but doesn't say what the class of the object is. `Vision` takes a `VisionIO` and when you construct it in `SwerveSubsystem.createRealCameras` you make a concrete class that implements `VisionIO` (`VisionIOReal`) so you're good. The default null implementation just makes it so if you make a `class Foo implements VisionIO` and you don't explicitly implement that method, it comes with a default implementation. That makes sense sometimes, but a null implementation is a foot-gun because it's easy to accidentally end up with a method that does nothing by default and the compiler doesn't force you to notice.

kevinclark: So long story short, cut the `default` and the `{}` and add a `;` at the end.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
kevinclark: Yeah, looks like they just reverted this change. Both a smaller queue and back to blocking.

Happy to talk in person or slack, but will try to give a summary here.

Queues have a few options when you try to put things in them. They can accept as much as you give or have a limit. When they hit a limit they can drop things on the floor, refuse to take them, or wait until they've got room and then add. `offer` is a nonblocking method by definition and returns `false` if it can't fit the item so we're giving the implementation an option to reject too-much-data.

So lets talk options we've seen.

 [ArrayDeque](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ArrayDeque.html) is unbounded ("no capacity restrictions; they grow as necessary) so `offer` will by definition never return `false`. That number you pass to the constructor is the initial capacity (so it allocates memory for 100 items all at once). This class is explicitly not thread safe, so you need to roll your own synchronization.

In contrast, [BlockingArrayQueue](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ArrayBlockingQueue.html) is defined with a *limit*. `offer` will return false if we try to go over the limit and it also has a `put` method (part of the [BlockingQueue](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/BlockingQueue.html) interface) which will pause the thread until there's room in the queue. That would be weird in a single-threaded environment but makes lots of sense if it's an extra worker thread. In any case, we're never going to block in this use because offer is being called. Effectively we've opted into limiting the size and dropping the data on the floor if we overrun. Honestly, I'm not sure that's the right choice either - probably we want the *most recent* data, not the first into the queue if we overrun. Probably a better choice is a circular buffer / ring buffer like the [EvictingQueue](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/EvictingQueue.html) implemented in Java which will drop the oldest data rather than the newest when it fills.

src/main/java/frc/robot/subsystems/vision/VisionIO.java
spellingcat: fixed!

kevinclark: You can click the "resolved" button and the comment chain will get archived.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: okay so from my understanding this is used when the thread is run. if it's false (if any of the signals are not CAN FD) it waits and then refreshes all signals and if it's true it just waits for new data, so yes it seems like it should be overwriting. however, if one call sets it as false and then a later call sets it as true before running the thread I don't know if it would handle that properly.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -409,6 +591,18 @@ public static Translation2d[] getModuleTranslations() {
...
 
+  public static VisionConstants[] getCameraConstants() {
+    return new VisionConstants[] {leftCamConstants, rightCamConstants};
+  }
+
+  /** Returns the module positions (turn angles and drive velocities) for all of the modules. */
+  @AutoLogOutput(key = "SwerveStates/Measured")
+  private SwerveModulePosition[] getModulePositions() {
+    SwerveModulePosition[] states = new SwerveModulePosition[4];
+    Arrays.stream(modules).map(Module::getPosition).toArray();
```
kevinclark, line 602: This line doesn't do anything because it's not stored. Probably need to be assigned to the var above? Or just returned outright?

kevinclark (changes requested):
Needs this one fix but otherwise should be good to go.

PR #35, opened 2024-02-07
remove io default impls and clean up mechanisms

Lewis-Seiden:
general cleanliness per kevins feedback in other mechanism prs

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -42,7 +42,14 @@ public class Robot extends LoggedRobot {
 
   private final SwerveSubsystem swerve =
       new SwerveSubsystem(
-          mode == RobotMode.REAL ? new GyroIOPigeon2() : new GyroIO() {},
+          mode == RobotMode.REAL ? new GyroIOPigeon2() : new GyroIO() {
+            // Blank implementation to mock gyro in sim
+            @Override
+            public void updateInputs(GyroIOInputs inputs) {}
+
+            @Override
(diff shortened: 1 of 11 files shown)


Lewis-Seiden:
forgot to pull main before creating this so the commit history is a bit messy

src/main/java/frc/robot/subsystems/shooter/ShooterIOSim.java
```diff
@@ -93,4 +90,9 @@ public void setFlywheelVoltage(final double left, final double right) {
     leftFlywheelSim.setInput(MathUtil.clamp(left, -12, 12));
     rightFlywheelSim.setInput(MathUtil.clamp(right, -12, 12));
   }
+
+  @Override
+  public void resetPivotPostion(Rotation2d rotation) {
```
kevinclark, line 95: Was this missing a bug?

kevinclark (approved):
Unclear to me that the updateInputs(outVariable) vs inputs = updateInputs() change is worth it for the `final`, but it looks like it brings it in line with other uses. No big deal either way.

Generally lgtm.

src/main/java/frc/robot/subsystems/shooter/ShooterIOSim.java
Lewis-Seiden: having it be a noop shouldve been fine? the point of the method is to reset the encoder to match physical hardware whereas this always has exact knowledge of its state so it shouldnt matter.

PR #39, opened 2024-02-10
aligned can and dio ids with spreadsheet

Lewis-Seiden:
https://docs.google.com/spreadsheets/d/1odhD145OL3dNaa40gRIr8u8jl-C9k7WLtFd2wuopsfE/edit?usp=sharing this spreadsheet


kevinclark (approved):
LGTM but didn't compare the spreadsheet to the numbers.

Some would argue that having a single file that maps from the numbers to constants that the individual files use would be easier to check against the spreadsheet (or better yet, replaces the spreadsheet entirely) but I don't feel particularly strongly about it. Having numbers in one place that strongly bind to the implementation seems fine. Just calling out other reasonable viewpoints.

Lewis-Seiden:
we did do that in previous years, which ended up with one very large and hard to look through file. this year im keeping each mechanism's constants in their relevant files. not convinced one way or the other at this point.

PR #44, opened 2024-02-16
Add elevator and shooter sysid commands

Lewis-Seiden:


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -417,14 +417,11 @@ public Command runModuleSteerCharacterizationCmd() {
     return Commands.sequence(
         this.runOnce(() -> SignalLogger.start()),
         moduleSteerRoutine.quasistatic(Direction.kForward),
-        this.stopCmd(),
-        Commands.waitSeconds(1.0),
+        this.stopCmd().withTimeout(1.0),
```
Awesomeplayer165, line 420: does this `1.0` want to be generic? I don't have a strong opinion on this as it probably won't change often, but if changed, then we'll need to change all of them..

Awesomeplayer165 (approved):
small things, don't have strong opinions on them, just coming from a different POV.

PR #48, opened 2024-02-23
Rework PhoenixOdometryThread

kevinclark:
This pulls all the thread-sensitive stuff inside PhoenixOdometryThread, makes it so we only pull each status signal once, and ensures that timestamps are 1:1 associated with a sampling.

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
+++ b/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
@@ -25,7 +25,7 @@ import com.ctre.phoenix6.hardware.TalonFX;
 import com.ctre.phoenix6.signals.FeedbackSensorSourceValue;
 import com.ctre.phoenix6.signals.InvertedValue;
 import com.ctre.phoenix6.signals.NeutralModeValue;
-import com.google.common.collect.Sets;
+import com.google.common.collect.ImmutableSet;
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.util.Units;
 import frc.robot.subsystems.swerve.Module.ModuleConstants;
@@ -157,8 +157,8 @@ public class ModuleIOReal implements ModuleIO {
 
     PhoenixOdometryThread.getInstance()
         .registerSignals(
-            new Registration(driveTalon, Sets.newHashSet(driveTalon.getPosition())),
-            new Registration(turnTalon, Sets.newHashSet(turnPosition)));
+            new Registration(driveTalon, ImmutableSet.of(drivePosition)),
+            new Registration(turnTalon, ImmutableSet.of(turnPosition)));
 
     BaseStatusSignal.setUpdateFrequencyForAll(
         Module.ODOMETRY_FREQUENCY_HZ, drivePosition, turnPosition);
@@ -202,7 +202,7 @@ public class ModuleIOReal implements ModuleIO {
 
     var samples =
         PhoenixOdometryThread.getInstance()
-            .samplesSince(lastUpdate, Sets.newHashSet(drivePosition, turnPosition));
+            .samplesSince(lastUpdate, ImmutableSet.of(drivePosition, turnPosition));
     lastUpdate = samples.get(samples.size() - 1).timestamp();
 
     inputs.odometryTimestamps = samples.stream().mapToDouble(s -> s.timestamp()).toArray();
--- a/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
+++ b/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
@@ -213,8 +213,9 @@ public class ModuleIOReal implements ModuleIO {
     inputs.odometryTurnPositions =
         samples.stream()
             // should be after offset + gear ratio
+            .map(s -> s.values().get(turnPosition))
             .filter(s -> s != null)
-            .map(s -> Rotation2d.fromRotations(s.values().get(turnPosition)))
+            .map(Rotation2d::fromRotations)
             .toArray(Rotation2d[]::new);
   }
 
--- a/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
+++ b/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
@@ -25,10 +25,11 @@ import com.ctre.phoenix6.hardware.TalonFX;
 import com.ctre.phoenix6.signals.FeedbackSensorSourceValue;
 import com.ctre.phoenix6.signals.InvertedValue;
 import com.ctre.phoenix6.signals.NeutralModeValue;
+import com.google.common.collect.Sets;
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.util.Units;
 import frc.robot.subsystems.swerve.Module.ModuleConstants;
-import java.util.Queue;
+import frc.robot.subsystems.swerve.PhoenixOdometryThread.Registration;
 
 /**
  * Module IO implementation for Talon FX drive motor controller, Talon FX turn motor controller, and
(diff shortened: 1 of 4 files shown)


kevinclark:
lastUpdate issue should be resolved

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,88 +59,97 @@ public static PhoenixOdometryThread getInstance() {
...
+      return journal.stream()
+          .filter(s -> s.timestamp > timestamp)
+          .map(
+              s -> {
+                var filteredValues =
+                    s.values.entrySet().stream()
+                        .filter(e -> signals.contains(e.getKey()))
+                        .collect(
+                            Collectors.toUnmodifiableMap(Map.Entry::getKey, Map.Entry::getValue));
+                return new Samples(s.timestamp, filteredValues);
```
Lewis-Seiden, line 113: any gc pressure concerns here?

src/test/java/frc/robot/subsystems/swerve/PhoenixOdometryThreadTest.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Queues;
+import frc.robot.subsystems.swerve.PhoenixOdometryThread.Samples;
+import java.util.Collections;
+import java.util.Map;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class PhoenixOdometryThreadTest {
```
Lewis-Seiden, line 16: not sure if we have time for it at this point in the season, but id love to go over unit testing and how to apply it to robot code at some point

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -11,19 +11,25 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 // GNU General Public License for more details.
 
+// Copyright 2024 FRC 8033 and Kevin Clark
```
Lewis-Seiden, line 14: since we dont do this in other files in the codebase i dont think we should do it here. it is probably worth putting a proper license for the project, since right now we just have the one that comes auto generated with wpilib projects

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
kevinclark: Not really. These are short lived objects that shouldn't make it out of the gc's eden. We're also limited to 20 samples stored in the evicting queue at the moment, so we're filtering over 20*signals.size objects total. Should be pretty slight in our use. Always better to measure than infer though, if we see issues.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
kevinclark: Yeah, the only reason I did it is that we have to keep the copyright notice for Mechanical Advantage since it's a derivative work, but it's so far removed from the original it's worth adding our names too. We could just do "Mechanical Advantage, Highlander Robotics, Kevin Clark" at the top line. Thoughts?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,88 +59,97 @@ public static PhoenixOdometryThread getInstance() {
...
-          // is provided by the documentation.
-          Thread.sleep((long) (1000.0 / Module.ODOMETRY_FREQUENCY_HZ));
-          if (signals.length > 0) BaseStatusSignal.refreshAll(signals);
-        }
-      } catch (InterruptedException e) {
-        e.printStackTrace();
-      } finally {
-        signalsLock.unlock();
-      }
+      var writeLock = journalLock.writeLock();
```
spellingcat, line 125: is there any specific reason why this isn't the same writeLock as before (just wondering since the old code had the same signalsLock)

spellingcat (commented):
i think i mostly get it but will probably still want to discuss on sunday. should we merge this into vision now/after sun, vision into main, and then changes from bring-up/bring-up-vision?

kevinclark:
I defer to @Lewis-Seiden on how you want to handle the merge. I'll be in tomorrow and we can go over the PR in depth. Will respond to questions posted momentarily.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -145,23 +144,22 @@ public ModuleIOReal(ModuleConstants constants) {
...
-    turnPositionQueue =
-        PhoenixOdometryThread.getInstance().registerSignal(turnTalon, turnTalon.getPosition());
     turnVelocity = turnTalon.getVelocity();
     turnAppliedVolts = turnTalon.getMotorVoltage();
     turnCurrent = turnTalon.getStatorCurrent();
 
+    PhoenixOdometryThread.getInstance()
+        .registerSignals(
+            new Registration(driveTalon, ImmutableSet.of(drivePosition)),
+            new Registration(turnTalon, ImmutableSet.of(turnPosition)));
```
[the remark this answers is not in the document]
kevinclark: So, I knew I wanted the api to support registering multiple signals to one parent device at once, so options are some time kind of [Collection](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html) or an Array. Arrays are pretty awkward to work with in Java most of the time, so unless I've got a specific reason to use them I don't by default. Of the major `Collection` types we're looking at `Set` and `List` since we don't need something with more complicated semantics like a `Queue` or a `Stack`. The major difference between `Set`s and `List`s is that `Set`s have unique elements while `List`s don't and `List`s have an order to them while `Set`s don't. The signals we're registering should be unique (no reason to register any of them more than once - in fact we should avoid this) and don't have an order. So `Set` it is. I'm passing data that shouldn't need to be modified (and won't stick around very long) so `ImmutableSet` is a reasonable choice. I could have as easily used `HashSet` - the signature for `registerSignals` only requires a `Set`, not a particular implementation of one.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -51,88 +59,97 @@ public static PhoenixOdometryThread getInstance() {
...
+    this.journal = journal;
   }
 
-  @Override
-  public void start() {
-    if (timestampQueues.size() > 0) {
-      super.start();
-    }
+  private PhoenixOdometryThread() {
+    this(EvictingQueue.create(20));
```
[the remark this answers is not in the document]
kevinclark: It's not, so I need to use external concurrency primatives (a read/write lock in this case) to make it so multiple threads can safely interact with it.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
kevinclark: A read/write lock is a kind of mutual exclusion construct that allows multiple readers, but only one writer. That means if we ever used this in an environment with more than one thread calling `samplesSince`, multiple callers could coexist in between journal writes. Once a write begins, all the readers would have to wait, but once the write was done they'd be able to coexist without issue. The other lock allows only one access at a time, regardless of what's going on.

kevinclark: In our current use it really doesn't make a difference. But it's a minor change that makes the class more future proof.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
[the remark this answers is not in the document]
spellingcat: that makes sense, I was mostly wondering why we need to support multiple signals per device since I don't believe it's ever implemented with multiple signals and i imagine that could cause issues?

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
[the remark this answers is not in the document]
spellingcat: i see, out of curiosity why aren't we wrapping it in the synchronized queue you mentioned earlier? is it because we need the locks either way?

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -202,17 +200,22 @@ public void updateInputs(ModuleIOInputs inputs) {
...
+    inputs.odometryTimestamps = samples.stream().mapToDouble(s -> s.timestamp()).toArray();
     inputs.odometryDrivePositionsMeters =
-        drivePositionQueue.stream().mapToDouble((Double value) -> value).toArray();
+        samples.stream().filter(s -> s != null).mapToDouble(s -> s.values().get(drivePosition)).toArray();
     inputs.odometryTurnPositions =
-        turnPositionQueue.stream()
-            .map(Rotation2d::fromRotations) // should be after offset + gear ratio
+        samples.stream()
+            // should be after offset + gear ratio
+            .filter(s -> s != null)
```
kevinclark, line 216: @Lewis-Seiden This (and the other filter) aren't quite right. Samples should never be null. The thing we're protecting from is that the call to .get might ask for a signal that wasn't collected. So we need to break the map into "map to the result of the get call, filter out the nulls, then map into the rotations".

Does that make sense?

PR #51, opened 2024-02-27
refresh inputs logging

Lyssia-Seiden:
adds supply current logging, standardizes on `getValueAsDouble()` instead of `getValue()`, makes sure to only use arrays where needed, and makes elevator properly log both motors.

### DIFF
--- a/src/main/java/frc/robot/utils/logging/TalonFXLogger.java
+++ b/src/main/java/frc/robot/utils/logging/TalonFXLogger.java
@@ -1,6 +1,7 @@
 package frc.robot.utils.logging;
 
 import com.ctre.phoenix6.BaseStatusSignal;
+import com.ctre.phoenix6.StatusCode;
 import com.ctre.phoenix6.StatusSignal;
 import com.ctre.phoenix6.hardware.TalonFX;
 import edu.wpi.first.util.struct.Struct;
@@ -22,7 +23,7 @@ public class TalonFXLogger {
     public double velocity = 0.0;
 
     // StatusCode error value. Currently just checks one signal.
-    public int errorCode = 0;
+    public StatusCode errorCode = StatusCode.StatusCodeNotInitialized;
 
     // Faults
     public boolean licenseFault = false;
@@ -34,7 +35,7 @@ public class TalonFXLogger {
         double temperatureCelsius,
         double position,
         double velocityRotationsPerSecond,
-        int errorCode,
+        StatusCode errorCode,
         boolean licenseFault) {
       this.appliedVolts = appliedVolts;
       this.statorCurrentAmps = statorCurrentAmps;
@@ -55,7 +56,7 @@ public class TalonFXLogger {
         double temperatureCelsius,
         double position,
         double velocityRotationsPerSecond,
-        int errorCode) {
+        StatusCode errorCode) {
       this(
           appliedVolts,
           statorCurrentAmps,
@@ -81,7 +82,7 @@ public class TalonFXLogger {
           temperatureCelsius,
           position,
           velocityRotationsPerSecond,
-          0,
+          StatusCode.StatusCodeNotInitialized,
           false);
     }
 
@@ -117,7 +118,7 @@ public class TalonFXLogger {
         double temp = bb.getDouble();
         double rotation = bb.getDouble();
         double velocity = bb.getDouble();
-        int errorCode = bb.getInt();
+        var errorCode = StatusCode.valueOf(bb.getInt());
         boolean licenseFault = bb.get() != 0;
         return new TalonFXLog(
             voltage, statorAmps, supplyAmps, temp, rotation, velocity, errorCode, licenseFault);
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/subsystems/shooter/ShooterIOReal.java
```diff
@@ -18,16 +18,23 @@ public class ShooterIOReal implements ShooterIO {
 
   private final StatusSignal<Double> pivotVelocity = pivotMotor.getVelocity();
   private final StatusSignal<Double> pivotVoltage = pivotMotor.getMotorVoltage();
-  private final StatusSignal<Double> pivotAmps = pivotMotor.getStatorCurrent();
+  private final StatusSignal<Double> pivotStatorCurrentAmps = pivotMotor.getStatorCurrent();
+  private final StatusSignal<Double> pivotSupplyCurrentAmps = pivotMotor.getSupplyCurrent();
   private final StatusSignal<Double> pivotTempC = pivotMotor.getDeviceTemp();
   private final StatusSignal<Double> pivotRotations = pivotMotor.getPosition();
```
kevinclark, line 24: This is repeated enough you should really make it a class. A record with a custom constructor that takes the motor would really clean this up trivially.

kevinclark (commented):
Nothing blocking, but I think the level of repetition (pulling out status signals, setting update timings) is enough that you should strongly consider making a proper class or record

src/main/java/frc/robot/subsystems/elevator/ElevatorIO.java
```diff
@@ -12,7 +12,8 @@ public static class ElevatorIOInputs {
     public double elevatorPositionMeters = 0.0;
     public double elevatorVelocityMetersPerSec = 0.0;
     public double elevatorAppliedVolts = 0.0;
-    public double[] elevatorCurrentAmps = new double[] {};
+    public double[] elevatorStatorCurrentAmps = new double[] {};
+    public double[] elevatorSupplyCurrentAmps = new double[] {};
```
Awesomeplayer165, line 16: Like Kevin's -- not blocking -- why are we using an array for this? In other languages, this is represented as a tuple, which is an indicator for abstracting to a class. (also see below relating comment).

src/main/java/frc/robot/subsystems/elevator/ElevatorSubsystem.java
```diff
@@ -91,7 +91,7 @@ public Command setExtensionCmd(DoubleSupplier meters) {
 
   public Command runCurrentZeroing() {
     return this.run(() -> io.setVoltage(-1.0))
-        .until(() -> inputs.elevatorCurrentAmps[0] > 40.0)
+        .until(() -> inputs.elevatorStatorCurrentAmps[0] > 40.0)
```
Awesomeplayer165, line 94: this syntax is very scary for me. Not sure if this is the syntax or norm for Java, but readers can be unsure of the value difference between index 0, 1, ... (even how many there are? Abstracting to a class like mentioned above should clear up your intentions). I always repeat the phrase: write once, read 10 times.

src/main/java/frc/robot/subsystems/elevator/ElevatorIOSim.java
```diff
@@ -39,7 +39,7 @@ public void updateInputs(final ElevatorIOInputsAutoLogged inputs) {
     inputs.elevatorPositionMeters = physicsSim.getPositionMeters();
     inputs.elevatorVelocityMetersPerSec = physicsSim.getVelocityMetersPerSecond();
     inputs.elevatorAppliedVolts = volts;
-    inputs.elevatorCurrentAmps = new double[] {physicsSim.getCurrentDrawAmps()};
+    inputs.elevatorStatorCurrentAmps = new double[] {physicsSim.getCurrentDrawAmps()};
```
Awesomeplayer165, line 42: sometimes using arrays can be confusing the total length. also since this length can be whole replaced by a new array (which looks like is what you are doing everytime -- constructing a new array instead of setting existing values).

src/main/java/frc/robot/subsystems/elevator/ElevatorIO.java
Awesomeplayer165: especially since this is not being used for iteration.

src/main/java/frc/robot/subsystems/elevator/ElevatorSubsystem.java
Awesomeplayer165: also see this line being repeated in `src/main/java/frc/robot/subsystems/shooter/ShooterSubystem.java` :151, but with a different syntax of using the array. maybe clearing some of this stuff up might help clear confusion among fresh eyes.

Awesomeplayer165 (commented):
Nothing blocking too. Might be helpful to look into array usecases and see if they can be abstracted to an array if they aren't being used for special array use cases (iteration).

EDIT: talked to @Lewis-Seiden, gonna break into two variables then adopt @kevinclark  suggestion.

src/main/java/frc/robot/subsystems/elevator/ElevatorSubsystem.java
kevinclark: Yeah, I agree. Seems like a good use for a record. Additionally, the value appears to be initialized to a 0 length array but then we do array access here that would be out of bounds if the value isn't set to something else. If the size of the array changes it strongly implies that maybe array isn't the right container.

Lewis-Seiden: the whole array use was a holdover from the akit examples, where they wanted you to use an array for all motor current and temperature since they consider motor count an implementation detail. i dont really agree w their view of it either, and will change it

PR #53, opened 2024-02-29
add trajectory files from dynamic-auto-visionless

Awesomeplayer165:


Awesomeplayer165:
I have included an advantagescope layout json file for auto testing in sim if needed.

PR #55, opened 2024-03-01
Fix odo thread

Lewis-Seiden:


Awesomeplayer165 (approved):

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -210,7 +210,7 @@ public void updateInputs(ModuleIOInputs inputs) {
     var samples =
         PhoenixOdometryThread.getInstance()
             .samplesSince(lastUpdate, ImmutableSet.of(drivePosition, turnPosition));
-    if (!samples.isEmpty()) {
+    if (samples.stream().filter(s -> s != null).count() != 0) {
```
kevinclark, line 213: I don't think this filter should ever apply. The journal never has null values pushed.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -121,16 +121,19 @@ public void run() {
...
-        BaseStatusSignal.waitForAll(
-            2.0 / Module.ODOMETRY_FREQUENCY_HZ, signals.toArray(new BaseStatusSignal[0]));
-        journal.add(
-            new Samples(timestampFor(signals), Maps.asMap(signals, s -> s.getValueAsDouble())));
+        var status =
+            BaseStatusSignal.waitForAll(
+                2.0 / Module.ODOMETRY_FREQUENCY_HZ, signals.toArray(new BaseStatusSignal[0]));
+        writeLock.lock();
+        if (status.isOK()) {
+          journal.add(
```
kevinclark, line 132: You could add the lock/try/finally in here since it's the only line it applies to.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -367,9 +367,16 @@ public void periodic() {
...
+    for (int i = 0; i < modules.length; i++) {
+      Logger.recordOutput(
+          "Swerve/Updates Since Last " + i, modules[i].getOdometryTimestamps().length);
+    }
     double[] sampleTimestamps =
         modules[0].getOdometryTimestamps(); // All signals are sampled together
-    int sampleCount = sampleTimestamps.length;
+    int sampleCount =
+        Math.min(
+            Arrays.stream(modules).mapToInt(m -> m.getOdometryTimestamps().length).min().getAsInt(),
```
kevinclark, line 378: What guarantee is there that two modules with the same number of results have the same sampled times? Do they need to?

Lewis-Seiden: they should as long as neither is taking the earliest element in the queue, since they should both get the same earliest sample and only use the samples both have.

kevinclark: What code makes them only use samples in common?

Lewis-Seiden: the more i think it through the more edge cases that could break it i think of, ill just do it properly

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -330,12 +329,13 @@ public void periodic() {
       camera.updateInputs();
       camera.processInputs();
     }
-    odometryLock.lock(); // Prevents odometry updates while reading data
+    var odometryReadLock = PhoenixOdometryThread.getInstance().getReadLock();
```
Lewis-Seiden, line 332: not sure how smelly this solution is, it is the lowest impact on the rest of the codebase though

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -367,9 +368,18 @@ public void periodic() {
...
 
   private void updateOdometry() {
+    for (int i = 0; i < modules.length; i++) {
+      Logger.recordOutput(
+          "Swerve/Updates Since Last " + i, modules[i].getOdometryTimestamps().length);
+    }
     double[] sampleTimestamps =
         modules[0].getOdometryTimestamps(); // All signals are sampled together
-    int sampleCount = sampleTimestamps.length;
+    int sampleCount =
```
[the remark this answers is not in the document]
kevinclark: Sketch for this:

```
// might make more sense to filter the samples than the modules after the samples are applied, but for reference
var commonTimestamps = Arrays.stream(modules).reduce(Sets.newHashSet(), (set, module) -> {
  set.addAll(module.getOdometryTimestamps()); // might need to map this to a proper value?
  return set;
}

// then filter each module's results based on whether the timestamp is in common
```

kevinclark (changes requested):
Couple of correctness things.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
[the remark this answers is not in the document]
kevinclark: Alternative thought - we're now checking for waitForAll's result before adding a sample. If it returns 'ok' we should be getting a complete list every time, right?  Each module should be guaranteed to get the same set, the only edge case being the first loop if we have multiple registrations and read in the middle? How much does that first read matter? Could maybe mitigate that by not starting the thread until everything is registered (which I think was on construction?).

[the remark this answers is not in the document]
Lewis-Seiden: first read shouldnt matter, and the robot should be stationary then anyways

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -116,28 +117,39 @@ public List<Samples> samplesSince(double timestamp, Set<StatusSignal<Double>> si
...
+          System.out.println("Odo thread error: " + status.toString());
+        }
       } finally {
         writeLock.unlock();
       }
     }
   }
 
-  private double timestampFor(Set<BaseStatusSignal> signals) {
+  private double timestampFor(Set<StatusSignal<Double>> signals) {
```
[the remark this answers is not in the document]
Lewis-Seiden: there was one place where we needed the specific type, and rather than make robust casting i decided to just change it to always use `StatusSignal<Double>` since thats all we use it for

PR #57, opened 2024-03-01
Odo update refactor

spellingcat:
this redoes the updateOdometry method to actually use the samples we're getting from the other thread, rather than what's passed through the inputs, hopefully makes setPose() actually work, and generally make things clearer.

seems to work in sim but will need to test on the bot once the odo thread is ready


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -540,7 +539,8 @@ public ChassisSpeeds getRobotRelativeSpeeds() {
   /** Returns the current odometry pose. */
   @AutoLogOutput(key = "Odometry/Robot")
   public Pose2d getPose() {
-    return estimator.getEstimatedPosition();
+    // return estimator.getEstimatedPosition(); //Vision
+    return pose;
```
Lewis-Seiden, line 543: do we want this change?

Lewis-Seiden (changes requested):
make sure to merge latest main in, maybe even post-sfr-cleanup since thats going into main soon

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: you would probably know that better than me

spellingcat:
looks like this worked on the robot today

PR #64, opened 2024-03-03
Static autos

Lewis-Seiden:


src/main/deploy/pathplanner/paths/Example Path.path
Awesomeplayer165, line 1: Is this necessary

Awesomeplayer165 (approved):
looks good so far

Lewis-Seiden:
autos likely need more tuning, but will hold off for now until we get to the field or have extra time

PR #66, opened 2024-03-06
loads field from deploy folder

spellingcat:
this is because we eventually want to be able to remap the field as the position of the tags shifts over time


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -266,6 +266,12 @@ public SwerveSubsystem(GyroIO gyroIO, VisionIO[] visionIOs, ModuleIO[] moduleIOs
                 (Measure<Voltage> volts) -> runDriveCharacterizationVolts(volts.in(Volts)),
                 null,
                 this));
+    try {
+      fieldTags =
+          new AprilTagFieldLayout(
+              Filesystem.getDeployDirectory().toPath().resolve("vision\2024-crescendo.json"));
+    } catch (Exception e) {
```
Lewis-Seiden, line 273: any way to handle if this fails, like print smth and load the default field?

Lewis-Seiden (changes requested):

Lewis-Seiden (approved):

PR #71, opened 2024-03-15
odometry thread gyro fix

spellingcat:
if any signals are missing, it adds the ones that are there instead of blocking all updates

(i don't think this would account for if it's one module that goes down instead of the gyro but i imagine if that's the case we have bigger things to worry about)


src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -136,7 +137,12 @@ public void run() {
           journal.add(
               new Samples(timestampFor(signals), Maps.asMap(signals, s -> s.getValueAsDouble())));
         } else {
-          // System.out.println("Odo thread error: " + status.toString());
+          journal.add(
+              new Samples(
+                  timestampFor(signals),
+                  signals.stream()
+                      .filter(s -> s.getStatus().equals(StatusCode.OK))
```
kevinclark, line 144: I believe `SwerveSubsystem::updateOdometry` relies on timestamps and sample data being the same size. I think you may run into array out of bounds errors if that's not the case.

kevinclark (changes requested):
I think some of the code assumes every sample will have all data (which is a bad assumption but I think it does it). I think this change is unsafe with that being the case and you'll need to resolve that as well. You could potentially write a unit test to show that's not the case easier than analyzing the code.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: i don't believe it does? `updateInputs()` for both the modules and gyro just stream `odometrySamples` straight to an array and the latest timestamp is set to whatever the last value of `odometrySamples` is (so, not reliant on what the size is between loops). outside of those two i don't think it's used

spellingcat: also it seemed fine when i ran it on the log

kevinclark: The problem is that the `sampleCount` is based on [the minimum number of odometryTimestamps in any module](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java#L381-L382) but `odometryTimestamp` is set [here](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java#L205-L208) and will be equal to the total number of samples because no filtering goes on. Also, note that `odometryTurnPostions` *does* filter out nulls (probably the right choice) and `odometryDrivePositions` doesn't (which means if you have a sample with no drive positions you're going to have arrays with nulls that you probably don't expect). Because you're doing some filtering, the number of turn positions *can* be smaller than the number of samples. [Right here](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/Module.java#L69) we index turn positions based on the number of samples and may see an out of bounds access. Looks like `updateOdometry` might actually get away with it when calling `getOdometryPositions` because the underlying array is based on sample counts, but it makes the [same mistake](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java#L415) with yaw which is filtered [here](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/GyroIOPigeon2.java#L53-L58).

Just want to be clear that the filtering isn't the underlying problem - you really *have* to do that if you don't know the sample is going to have all signals at all times. The problem is the assumption that the number of timestamps and the number of sample data points will be the same.

kevinclark (changes requested):
Unfortunately the assumption that len(timestamps) == len(signal data points) is an out of bound error waiting to happen.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
spellingcat: i'm sorry if i'm still not understanding but i believe `odometryTimestamps` gets it from the `asyncOdometrySamples ` passed into `updateInputs`, [which should be `samplesSince()` in `PhoenixOdometryThread`](https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java#L334-L342), which should be the journal which has already been filtered. i think the same goes for turn positions, drive positions, and yaw positions.

do you think we should filter out nulls for `odometryDrivePositions` as well?

src/main/java/frc/robot/subsystems/swerve/GyroIOPigeon2.java
```diff
@@ -53,8 +55,8 @@ public void updateInputs(GyroIOInputs inputs, List<Samples> asyncOdometrySamples
     inputs.odometryYawPositions =
         asyncOdometrySamples.stream()
             .map(s -> s.values().get(yaw))
-            .map(d -> d == null ? null : Rotation2d.fromDegrees(d))
-            .toArray(Rotation2d[]::new);
+            .map(d -> d == null ? new NullableRotation2d(null) : new NullableRotation2d(new Rotation2d(d)))
```
kevinclark, line 58: Does this still need to be a Rotation2d.fromDegrees? A static factory on NullableRotation2d might make this a little nicer. `NullableRoation2d.fromDegrees(d)`?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -407,11 +407,11 @@ private void updateOdometry() {
       // sample in x, y, and theta based only on the modules, without
       // the gyro. The gyro is always disconnected in simulation.
       Twist2d twist = kinematics.toTwist2d(modulePositions);
-      if (!gyroInputs.connected || gyroInputs.odometryYawPositions[deltaIndex] == null) {
+      if (!gyroInputs.connected || gyroInputs.odometryYawPositions[deltaIndex].get() == null) {
         // We don't have a complete set of data, so just use the module rotations
         rawGyroRotation = rawGyroRotation.plus(new Rotation2d(twist.dtheta));
```
kevinclark, line 412: Might be worth logging when we see this.

src/main/java/frc/robot/utils/NullableRotation2d.java
```diff
@@ -0,0 +1,26 @@
...
+
+package frc.robot.utils;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.util.struct.StructSerializable;
+
+/** Wrapper over WPILib's Rotation2d to allow serialization of nulls.*/
+public class NullableRotation2d implements StructSerializable {
+    boolean isNull;
+    Rotation2d value;
```
kevinclark, line 13: I'd final these.

src/main/java/frc/robot/subsystems/swerve/Module.java
```diff
@@ -66,8 +67,12 @@ public void periodic() {
     odometryPositions = new SwerveModulePosition[sampleCount];
     for (int i = 0; i < sampleCount; i++) {
       double positionMeters = inputs.odometryDrivePositionsMeters[i];
-      Rotation2d angle = inputs.odometryTurnPositions[i];
-      odometryPositions[i] = new SwerveModulePosition(positionMeters - lastPositionMeters, angle);
+      NullableRotation2d angle = inputs.odometryTurnPositions[i];
+      if (angle.get() == null) {
+        odometryPositions[i] = null; // SwerveSubsystem deals with this
```
kevinclark, line 72: Where does it deal with this? I see it doing a check on odometryYawPositions but not odometryPositions.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -210,8 +212,14 @@ public void updateInputs(final ModuleIOInputs inputs, final List<Samples> asyncO
...
-            .map(d -> d == null ? null : Rotation2d.fromRotations(d))
-            .toArray(Rotation2d[]::new);
+            .map(d -> d == null ? new NullableRotation2d(null) : new NullableRotation2d(Rotation2d.fromRotations(d)))
+            .toArray(NullableRotation2d[]::new);
+
+        //     inputs.odometryYawPositions =
+        // asyncOdometrySamples.stream()
+        //     .map(s -> s.values().get(yaw))
+        //     .map(d -> d == null ? new NullableRotation2d(true) : new NullableRotation2d(new Rotation2d(d)))
+        //     .toArray(NullableRotation2d[]::new);
```
kevinclark, line 222: Might as well just delete it?

src/main/java/frc/robot/subsystems/swerve/ModuleIO.java
```diff
@@ -33,7 +34,7 @@ public static class ModuleIOInputs {
     public double[] turnCurrentAmps = new double[] {};
 
     public double[] odometryDrivePositionsMeters = new double[] {};
-    public Rotation2d[] odometryTurnPositions = new Rotation2d[] {};
+    public NullableRotation2d[] odometryTurnPositions = new NullableRotation2d[] {};
```
kevinclark, line 37: I also don't see it handling null odometryTurnPositions.

kevinclark (changes requested):
Think a couple usages were missed?

src/main/java/frc/robot/subsystems/swerve/Module.java
spellingcat: https://github.com/HighlanderRobotics/Crescendo/blob/gyro-odo-fix/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java#L388-L391

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
spellingcat: oops

src/main/java/frc/robot/subsystems/swerve/ModuleIO.java
spellingcat: that should be handled on L70

src/main/java/frc/robot/Robot.java
```diff
@@ -73,7 +73,7 @@ public static enum Target {
     SPEAKER
   }
 
-  public static final RobotMode mode = Robot.isReal() ? RobotMode.REAL : RobotMode.SIM;
+  public static final RobotMode mode = Robot.isReal() ? RobotMode.REAL : RobotMode.REPLAY;
```
kevinclark, line 76: Is this part of the gyro fix? What's the expectation here?

src/main/java/frc/robot/subsystems/swerve/GyroIOPigeon2.java
```diff
@@ -53,9 +54,8 @@ public void updateInputs(GyroIOInputs inputs, List<Samples> asyncOdometrySamples
     inputs.odometryYawPositions =
         asyncOdometrySamples.stream()
             .map(s -> s.values().get(yaw))
-            .filter(s -> s != null)
-            .map(Rotation2d::fromDegrees)
-            .toArray(Rotation2d[]::new);
+            .map(d -> d == null ? new NullableRotation2d(null) : NullableRotation2d.fromDegrees(d))
```
kevinclark, line 57: Minor usage improvement - you might consider adding a helper that does this. Think an analog to Optional.ofNullable. The only awkwardness is the degrees/rotations question. `ofNullableDegrees`?

Or not. Not a major change.

src/main/java/frc/robot/subsystems/swerve/Module.java
```diff
@@ -131,47 +133,45 @@ public void stop() {
     io.setDriveVoltage(0.0);
   }
 
-  /** Returns the current turn angle of the module. */
+  /** Returns the current turn angle of the module without PhoenixOdometryThread. */
```
kevinclark, line 136: Might make more sense to say where this data comes from? As it's written I'm not actually sure.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -220,9 +221,12 @@ public void updateInputs(final ModuleIOInputs inputs, final List<Samples> asyncO
...
             // should be after offset + gear ratio
             .map(s -> s.values().get(turnPosition))
-            .filter(s -> s != null)
-            .map(Rotation2d::fromRotations)
-            .toArray(Rotation2d[]::new);
+            .map(
+                d ->
+                    d == null
+                        ? new NullableRotation2d(null)
+                        : new NullableRotation2d(Rotation2d.fromRotations(d)))
```
kevinclark, line 228: Same note as the other usage. Not a big deal.

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -127,17 +127,19 @@ public void run() {
       var writeLock = journalLock.writeLock();
       // NOTE (kevinclark): The toArray here in a tight loop is kind of ugly
       // but keeping up a symmetric array is too and it's probably negligible on latency.
-      var status =
```
kevinclark, line 130: You might consider still reporting if status is coming back false. Seems useful for diagnostics.

kevinclark: Maybe per signal down below?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -476,7 +478,7 @@ private void updateVision() {
         camera.setSimPose(estPose, camera, newResult);
         Logger.recordOutput("Vision/Vision Pose From " + camera.getName(), visionPose);
         Logger.recordOutput("Vision/Vision Pose2d From " + camera.getName(), visionPose.toPose2d());
-        pose = pose.interpolate(visionPose.toPose2d(), 0.25);
+        // pose = pose.interpolate(visionPose.toPose2d(), 0.25);
```
kevinclark, line 481: Delete vs comment out? Do we never want to interpolate the pose?

src/main/java/frc/robot/utils/NullableRotation2d.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.util.struct.StructSerializable;
+
+/** Wrapper over WPILib's Rotation2d to allow serialization of nulls. */
+public class NullableRotation2d implements StructSerializable {
+  final boolean isNull;
+  final Rotation2d value;
+
+  public NullableRotation2d(Rotation2d value) {
+    isNull = value == null ? true : false;
```
kevinclark, line 16: When you find yourself saying: `someCondition ? true : false` you can just use `someCondition`.

So, `isNull = value == null`.

src/main/java/frc/robot/subsystems/swerve/GyroIOPigeon2.java
kevinclark: Ah, I see you've already got a `fromDegrees` but it doesn't take a `Double` so it can't be null. You could pretty easily repurpose that to do this.

kevinclark (commented):
This looks great! Very minor changes suggested but not blocking.

src/main/java/frc/robot/Robot.java
spellingcat: no this just gets changed back and forth depending on if we want to run it in sim or replay it in the log

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
spellingcat: from my understanding this was a temp fix before we realized that the timestamps were not synced. think this can be deleted

src/main/java/frc/robot/Robot.java
kevinclark: Seems like a weird thing to commit then?

src/main/java/frc/robot/subsystems/swerve/Module.java
spellingcat: I think it's just the not high freq samples

kevinclark (approved):

PR #76, opened 2024-04-08
Manual overrides

Lewis-Seiden:


src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -88,14 +90,23 @@ public Command setRunAlongCmd(
...
-            Commands.either(
-                this.setBlinkingCmd(new Color("#ffff00"), new Color(), 10.0)
-                    .until(() -> !targetIsSpeaker.getAsBoolean() || !enabled.getAsBoolean()),
-                this.setBlinkingCmd(new Color("#ff7777"), new Color(), 10.0)
-                    .until(() -> targetIsSpeaker.getAsBoolean() || !enabled.getAsBoolean()),
-                targetIsSpeaker),
+            Commands.select(
+                Map.of(
+                    Target.SPEAKER,
+                    this.setBlinkingCmd(new Color("#ffff00"), new Color(), 10.0)
```
Awesomeplayer165, line 98: good colors!

Awesomeplayer165 (changes requested):
small things, functionality should be fine. make sure to test it out.

src/main/java/frc/robot/Robot.java
```diff
@@ -187,7 +189,13 @@ public void robotInit() {
...
     feeder.setDefaultCommand(
         Commands.repeatingSequence(
-            feeder.indexCmd().until(() -> currentTarget != Target.SPEAKER),
+            feeder
+                .indexCmd()
+                .until(
+                    () ->
+                        currentTarget != Target.SPEAKER
+                            || currentTarget != Target.FEED
+                            || currentTarget != Target.SUBWOOFER),
```
kevinclark, line 198: Having a little trouble following this. What's the idea here?

Lewis-Seiden: the feeder should run indexing until the target is changed to be something that isnt one of the speaker modes. right now thats just the amp, and it could be refactored to reflect that. it was more readable before this refactor, see if the refactor around jacobs comment makes it better

kevinclark: So should this be && instead of ||? Right now this always evaluates to `true`, doesn't it?

src/main/java/frc/robot/Robot.java
```diff
@@ -258,17 +263,37 @@ public void robotInit() {
...
+                    .asProxy(),
+                shooter.runStateCmd(
+                    AutoAim.FEED_SHOT.getRotation(),
+                    AutoAim.FEED_SHOT.getLeftRPS(),
+                    AutoAim.FEED_SHOT.getRightRPS())))
         .onFalse(
             Commands.parallel(
                     shooter.run(() -> {}), feeder.runVelocityCmd(FeederSubsystem.INDEXING_VELOCITY))
                 .withTimeout(0.5));
+    controller
```
kevinclark, line 280: Feels a little like this can be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)'d up since the command appears to be identical aside from the target. Not a strong push, but I worry a bit that you tune one and forget to touch the others. Unless you think they're actually independent?

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
kevinclark: Love the visual confirmation of what target has been selected 👍

kevinclark (approved):

src/main/java/frc/robot/Robot.java
Lewis-Seiden: there isnt a great way to abstract the trigger bindings here that doesnt result in more verbosity. it isnt inconceivable that we want these to have more different behaviour later either.

PR #77, opened 2024-04-11
Handle null modules

spellingcat:
added back nullable doubles, it should be able to survive module and gyro failures now and it will update the member variable pose with the gyro data if it has that but no modules. i would like to also update the estimator w/only gyro but i don't think that's super urgent rn or even possible

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
+++ b/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
@@ -33,6 +33,7 @@ import edu.wpi.first.math.util.Units;
 import frc.robot.subsystems.swerve.Module.ModuleConstants;
 import frc.robot.subsystems.swerve.PhoenixOdometryThread.Registration;
 import frc.robot.subsystems.swerve.PhoenixOdometryThread.Samples;
+import frc.robot.utils.NullableDouble;
 import frc.robot.utils.NullableRotation2d;
 import java.util.List;
 
@@ -216,7 +217,11 @@ public class ModuleIOReal implements ModuleIO {
     inputs.odometryTimestamps =
         asyncOdometrySamples.stream().mapToDouble(s -> s.timestamp()).toArray();
     inputs.odometryDrivePositionsMeters =
-        asyncOdometrySamples.stream().mapToDouble(s -> s.values().get(drivePosition)).toArray();
+        asyncOdometrySamples.stream()
+            .map(s -> s.values().get(drivePosition))
+            .map(d -> d == null ? new NullableDouble(null) : new NullableDouble(d))
+            .toArray(NullableDouble[]::new);
+    // mapToDouble(s -> s.values().get(drivePosition)).toArray();
     inputs.odometryTurnPositions =
         asyncOdometrySamples.stream()
             // should be after offset + gear ratio
--- a/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
+++ b/src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
@@ -219,9 +219,8 @@ public class ModuleIOReal implements ModuleIO {
     inputs.odometryDrivePositionsMeters =
         asyncOdometrySamples.stream()
             .map(s -> s.values().get(drivePosition))
-            .map(d -> d == null ? new NullableDouble(null) : new NullableDouble(d))
+            .map(NullableDouble::new)
             .toArray(NullableDouble[]::new);
-    // mapToDouble(s -> s.values().get(drivePosition)).toArray();
     inputs.odometryTurnPositions =
         asyncOdometrySamples.stream()
             // should be after offset + gear ratio
--- a/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
@@ -429,12 +429,14 @@ public class SwerveSubsystem extends SubsystemBase {
       SwerveModulePosition[] modulePositions = new SwerveModulePosition[4];
       SwerveModulePosition[] moduleDeltas =
           new SwerveModulePosition[4]; // change in positions since the last update
+      boolean isNull = false;
       for (int moduleIndex = 0; moduleIndex < 4; moduleIndex++) {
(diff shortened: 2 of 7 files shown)


src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -216,7 +217,11 @@ public void updateInputs(final ModuleIOInputs inputs, final List<Samples> asyncO
     inputs.odometryTimestamps =
         asyncOdometrySamples.stream().mapToDouble(s -> s.timestamp()).toArray();
     inputs.odometryDrivePositionsMeters =
-        asyncOdometrySamples.stream().mapToDouble(s -> s.values().get(drivePosition)).toArray();
+        asyncOdometrySamples.stream()
+            .map(s -> s.values().get(drivePosition))
+            .map(d -> d == null ? new NullableDouble(null) : new NullableDouble(d))
```
kevinclark, line 222: You're always passing d to the constructor. Might as well just `.map(NullableDouble::new)`.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -216,7 +217,11 @@ public void updateInputs(final ModuleIOInputs inputs, final List<Samples> asyncO
...
+            .toArray(NullableDouble[]::new);
+    // mapToDouble(s -> s.values().get(drivePosition)).toArray();
```
kevinclark, line 224: Delete

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -429,19 +429,41 @@ private void updateOdometry() {
       SwerveModulePosition[] modulePositions = new SwerveModulePosition[4];
       SwerveModulePosition[] moduleDeltas =
           new SwerveModulePosition[4]; // change in positions since the last update
+      boolean isNull = false;
```
kevinclark, line 432: Better name will make this flow better. `hasNullModulePosition`?

src/main/java/frc/robot/utils/NullableDouble.java
```diff
@@ -0,0 +1,24 @@
...
+
+import edu.wpi.first.util.struct.StructSerializable;
+
+/** Silly */
+public class NullableDouble implements StructSerializable {
+  final boolean isNull;
+  final Double value;
+
+  public NullableDouble(Double value) {
+    isNull = value == null ? true : false;
```
kevinclark, line 15: `isNull = value == null`

src/main/java/frc/robot/utils/NullableDouble.java
```diff
@@ -0,0 +1,24 @@
...
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.utils;
+
+import edu.wpi.first.util.struct.StructSerializable;
+
+/** Silly */
+public class NullableDouble implements StructSerializable {
+  final boolean isNull;
+  final Double value;
```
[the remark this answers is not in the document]
kevinclark: @Lewis-Seiden Whole point of the class is to allow null Doubles to be serialized. You could certainly use a `double` and set it to an arbitrary value. Might as well just use the boxed version though.

@spellingcat It *does* mean we can get away without actually storing isNull though. It's just needed for serialization. You can do the same thing with NullableRotation2d now that I think about it.

kevinclark (changes requested):
Minor changes noted. There's nothing that would keep this from working if we were in a rush but a couple small tweaks will make this nicer.

src/main/java/frc/robot/utils/NullableDouble.java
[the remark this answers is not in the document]
spellingcat: i think i made it a Double because in Module.java it checks if .get() == null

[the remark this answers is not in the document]
kevinclark: Since it's returning null if the value is null, and otherwise the value, you can just return the value. is null is really only something that's needed in serialization. You can cut the variable here and just use it in the struct serialization implementation.

src/main/java/frc/robot/utils/NullableDoubleStruct.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public String getSchema() {
+    return "boolean isNull; double value;";
+  }
+
+  @Override
+  public NullableDouble unpack(ByteBuffer bb) {
+    byte isNull = bb.get();
+    double value = bb.getDouble();
+    if (isNull == 0) {
```
kevinclark, line 36: Minor note - 1 is typically used for true. You're using it consistently here, so it's not a big deal. Little odd though.

kevinclark (commented):
Might as well simplify NullableDouble by removing isNull (which is redundant). Not blocking though.

src/main/java/frc/robot/utils/NullableDoubleStruct.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public String getSchema() {
+    return "boolean isNull; double value;";
+  }
+
+  @Override
+  public NullableDouble unpack(ByteBuffer bb) {
+    byte isNull = bb.get();
+    double value = bb.getDouble();
+    if (isNull == -1) {
```
kevinclark, line 36: So close! Nuke the negative.

src/main/java/frc/robot/utils/NullableDoubleStruct.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  @Override
+  public void pack(ByteBuffer bb, NullableDouble value) {
+    if (value == null) {
+      bb.put((byte) -1);
+      bb.putDouble(0);
+    } else {
+      bb.put((byte) 1);
```
kevinclark, line 49: And make this 0

kevinclark (changes requested):
The -1 is weirder. Either make it positive or just revert this. The last version was really fine. We're in solid nitpick territory.

spellingcat:
lol i read the "Minor note - 1" as negative one

kevinclark (approved):
👍👍👍

PR #82, opened 2024-09-07
Fix async odometry to use an IO layer

Lewis-Seiden:
not marked as merging to main bc this was branched off of traj-following-tuning, but that should be able to be merged as soon as its tested during drive practice to double check current limit changes.

This branch makes the odometry thread return a map of signal type (drive, steer, or gyro) and module numerical id to values, instead of status signals to values. This is so that the return value of samplesSince is serializable so that it can be plumbed through an IO layer, to fix replay on the async odo thread. The end goal of this is to connect the asynchronous odometry to simulation. This PR also removes filtering by signal for samplesSince, since they are all logged together anyways.


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -408,37 +411,31 @@ public void periodic() {
...
+              new SwerveModulePosition(
+                  sample.values().get(new SignalID(SignalType.DRIVE, moduleIndex)),
+                  Rotation2d.fromRotations(
+                      sample
+                          .values()
+                          .get(
+                              new SignalID(
+                                  SignalType.STEER,
+                                  moduleIndex)))); // gets positions from the thread, NOT inputs
+        } catch (NullPointerException e) {
```
kevinclark, line 438: This catch is overly broad and I'd suggest you instead handle nulls where they may occur. The problem with just catching the NPE is that you could have NPEs which are the result of a mistake of the code vs expected null values being passed and now you've papered over them like they're the same thing. There be dragons here.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -453,17 +450,18 @@ private void updateOdometry() {
         lastModulePositions[moduleIndex] = modulePositions[moduleIndex];
       }
       if (hasNullModulePosition) {
-        if (!gyroInputs.connected || gyroInputs.odometryYawPositions[deltaIndex].get() == null) {
+        if (!gyroInputs.connected
+            || sample.values().get(new SignalID(SignalType.GYRO, -1)) == null) {
```
kevinclark, line 454: Hardcoding of the gyro signal id has the potential to drift. If you're going to do it, I'd suggest a constant that gets reused.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -472,22 +470,22 @@ private void updateOdometry() {
       // sample in x, y, and theta based only on the modules, without
       // the gyro. The gyro is always disconnected in simulation.
       Twist2d twist = kinematics.toTwist2d(moduleDeltas);
-      if (!gyroInputs.connected || gyroInputs.odometryYawPositions[deltaIndex].get() == null) {
+      if (!gyroInputs.connected || sample.values().get(new SignalID(SignalType.GYRO, -1)) == null) {
```
kevinclark, line 473: Same as above.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -472,22 +470,22 @@ private void updateOdometry() {
...
         Logger.recordOutput("Odometry/Received Gyro Update", false);
         // We don't have a complete set of data, so just use the module rotations
         rawGyroRotation = rawGyroRotation.plus(new Rotation2d(twist.dtheta));
       } else {
         Logger.recordOutput("Odometry/Received Gyro Update", true);
-        rawGyroRotation = gyroInputs.odometryYawPositions[deltaIndex].get();
+        rawGyroRotation =
+            Rotation2d.fromDegrees(sample.values().get(new SignalID(SignalType.GYRO, -1)));
```
kevinclark, line 480: And here.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -794,13 +792,11 @@ public Rotation2d getRotation() {
 
   /** Sets the current odometry pose. */
   public void setPose(Pose2d pose) {
-    this.pose = pose;
     try {
       estimator.resetPosition(lastGyroRotation, getModulePositions(), pose);
     } catch (Exception e) {
       System.out.println("odo reset sad :(");
```
kevinclark, line 798: Should anything else happen here? Should we at least log the exception and know *why* it's sad?

src/test/java/frc/robot/subsystems/swerve/PhoenixOdometryThreadTest.java
```diff
@@ -42,50 +42,40 @@ void statusSignalsFromDifferentDevicesShouldNotBeEqual() {
...
     var thread = PhoenixOdometryThread.createWithJournal(Queues.newArrayDeque());
-    var samples = thread.samplesSince(0, ImmutableSet.of(talon.getAcceleration()));
+    var samples = thread.samplesSince(0.0);
     Assertions.assertEquals(Collections.emptyList(), samples);
   }
 
   @Test
   void samplesSinceReturnsAfterTimestamp() {
-    Map<StatusSignal<Double>, Double> sampleValue = ImmutableMap.of(talon.getAcceleration(), 42.0);
+    Map<SignalID, Double> sampleValue = ImmutableMap.of(new SignalID(SignalType.GYRO, -1), 42.0);
```
kevinclark, line 51: Hardcoded ID again and down the file.

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
```diff
@@ -224,24 +232,6 @@ public void updateInputs(final ModuleIOInputs inputs, final List<Samples> asyncO
...
-            .toArray(NullableDouble[]::new);
-    inputs.odometryTurnPositions =
-        asyncOdometrySamples.stream()
-            // should be after offset + gear ratio
-            .map(s -> s.values().get(turnPosition))
-            .map(
-                d ->
-                    d == null
-                        ? new NullableRotation2d(null)
-                        : new NullableRotation2d(Rotation2d.fromRotations(d)))
```
kevinclark, line 243: Should we just cut the nullable wrappers? Is the assumption that we just won't see the result vs supporting an explicit null result? You explicitly don't handle the case where we have fewer than 4 modules returning valid data - have we tested the case where that happens or where the gyro gets disconnected?

kevinclark (changes requested):
I think this is broadly in the right direction but have some questions about correctness and a few suggested design tweaks. All in all nice and considered work.

src/main/java/frc/robot/subsystems/swerve/OdometryThreadIO.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public void toLog(LogTable table) {
+      var timestamps = sampledStates.stream().mapToDouble(Samples::timestamp).toArray();
+      Set<Integer> modIds = new HashSet();
+      for (int i = 0; i < sampledStates.size(); i++) {
+        var sample = sampledStates.get(i);
+        for (var signal : sample.values().entrySet()) {
+          table.put(
+              "Data/" + i + " " + signal.getKey().type().toString() + " " + signal.getKey().modID(),
```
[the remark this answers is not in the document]
Lewis-Seiden: im using the index so that each timestamp has some identifier to get all the timestamps at that system. originally i put the timestamp in the key, but that means a new NT topic is created for each timestamp sample which floods ascope with topics. I like having the index before the modID so that all the values from one timestamp are next to each other for comparison for debugging. I think the order should be preserved since this order comes from `journal`s order in `PhoenixOdometryThread`, which I think should be constant since the thread is always sampled in time order? that contract isnt explicit at the moment though (although the only other user, the sim on the maple-swerve-sim branch, also maintains it)

src/main/java/frc/robot/subsystems/swerve/PhoenixOdometryThread.java
```diff
@@ -38,14 +42,36 @@
...
 
-  public record Samples(double timestamp, Map<StatusSignal<Double>, Double> values) {}
+  public record Samples(double timestamp, Map<SignalID, Double> values) {}
 
   private final ReadWriteLock journalLock = new ReentrantReadWriteLock(true);
 
-  private final Set<StatusSignal<Double>> signals = Sets.newHashSet();
+  // For gyros
+  private static final ModuleConstants NEGATIVE_ONE =
+      new ModuleConstants(-1, "", -1, -1, -1, Rotation2d.fromRotations(0));
```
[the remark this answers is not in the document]
Lewis-Seiden: we do use them as integers for indexing the module array as part of `updateOdometry`. It would be trivial to cast a `String` there, but i feel like casting is inviting errors vs just having a constant for gyros to special case that signal.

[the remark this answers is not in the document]
Lewis-Seiden: missing rust enums for some of this, would be nice to have `DRIVE` and `STEER` contain an id while `GYRO` doesnt

src/main/java/frc/robot/subsystems/swerve/ModuleIOReal.java
Lewis-Seiden: have not tested that yet, which is important. will try to get to it on sunday. in that case I believe `updateOdometry` should see a null module position or gyro angle and handle it.

kevinclark (approved):
Minor note about another way to handle the optional check. Otherwise looks good. I *think* at this point the nullable wrappers are unused and can be deleted, but up to you to confirm and non-blocking.

PR #92, opened 2024-11-07
Add per-module trajectory force feedforward to Choreo path following

Lyssia-Seiden:


Lyssia-Seiden:
this has not been tested on the robot, but its going to be a pain to get a field setup for testing and i want to transfer this over to 2025 code so im going to just merge it. this should be fine given we're just running random dev branches or `demo-mode` on banshee atp

src/main/java/frc/robot/subsystems/shooter/ShooterIOSim.java
```diff
@@ -35,7 +35,7 @@ public class ShooterIOSim implements ShooterIO {
       new DCMotorSim(DCMotor.getKrakenX60Foc(1), ShooterSubsystem.FLYWHEEL_RATIO, 0.001);
 
   ProfiledPIDController pivotController =
-      new ProfiledPIDController(1.0, 0.0, 1.0, new Constraints(10.0, 10.0));
+      new ProfiledPIDController(100.0, 0.0, 1.0, new Constraints(10.0, 10.0));
```
kevinclark, line 38: Holy kP Batman! I know this is sim, but dang. Is this on purpose?

src/main/java/frc/robot/subsystems/swerve/Module.java
```diff
@@ -64,24 +64,27 @@ public void periodic() {
 
   /** Runs the module closed loop with the specified setpoint state. Returns the optimized state. */
   public SwerveModuleState runSetpoint(SwerveModuleState state) {
-    final var optimizedState = SwerveModuleState.optimize(state, getAngle());
-    return runSetpoint(
-        state,
-        (optimizedState.speedMetersPerSecond - lastSetpoint.speedMetersPerSecond)
-            / (Timer.getFPGATimestamp() - lastTime));
+    return runSetpoint(state, 0.0, 0.0);
```
kevinclark, line 67: We mean to be running this with no force by default? Looks `accel` *had* previously had a value.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -518,6 +499,24 @@ private void updateOdometry() {
     }
   }
 
+  /** Generates a set of samples without using the async thread */
```
kevinclark, line 502: Is there a reason not to reuse the results from the thread? Looks like they're not being waited on anyway, so they should be similarly up to date?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -554,16 +562,32 @@ private void runVelocity(ChassisSpeeds speeds) {
...
+    }
 
     // Log setpoint states
     Logger.recordOutput("SwerveStates/Setpoints", setpointStates);
+    Logger.recordOutput("SwerveStates/ForceSetpoints", forceSetpoints);
     Logger.recordOutput("SwerveStates/SetpointsOptimized", optimizedSetpointStates);
   }
 
+  private void runVelocity(ChassisSpeeds speeds) {
+    // do the allocations here matter?
```
kevinclark, line 587: If a 64 byte allocation is a problem, we're doing something right everywhere else.

kevinclark (approved):
Seems like it makes sense on face aside from the note on Module.java. Would love to see some actual testing, but I understand if you're just getting this down for later.

src/main/java/frc/robot/subsystems/shooter/ShooterIOSim.java
Lyssia-Seiden: it is on purpose. given that the sim doesnt have the same physical constraints, something which just gets to the target was desirable (prior behaviour was a very very slow final approach) this sim is mainly used in auto routines whose simulations are mainly for logic and drivetrain motion, not the specific physics of this mechanism. next season might want to pay more attention to setting sims like this up for timing validation, but given that this codebase is being retired its fine.

src/main/java/frc/robot/subsystems/swerve/Module.java
Lyssia-Seiden: the old version calculated a guess for our acceleration target by comparing our current and last velocity setpoint. i have a hunch that this was noisier than it was helpful, but the big issue is the amount of force required varies depending on if the module is trying to linearly accelerate the robot or rotate the robot (motion is resisted by the robot's mass vs moment of inertia). this made the previous acceleration setpoint a poor model. this version directly takes force setpoints from choreo to account for this. however we do not currently have math implemented to handle generating those force setpoints outside of choreo. kP *should* be enough to handle this in most cases where precise acceleration is less of an issue, so having that be the default seems sane to me. having a default of 0 also should make it easier to debug force feedforwards since its easy to turn them off (i would be surprised if there isnt some issue when first running this).

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
Lyssia-Seiden: this is mainly for the sim, where the thread doesn't run. it would also kick in if the thread fails for some reason. on the 2025 codebase the robot will log an `Alert` if this is in use instead of the thread, so that the latter case is easier to find and fix.

